Issue #16, opened 2016-02-05
Re-queue multiple jobs


dfaller:
Improve the REST API to allow re-queuing multiple jobs in one REST call.
Include an option to reset the jobs to a new priority.

dfaller:
Removed the canceling of multiple jobs from this issue. Scaling back the scope to just re-queuing.

dfaller:
This is now ready for the new REST API call(s). I think there are two use cases here:
1. Re-queuing a job or jobs by ID
2. Re-queuing a bunch of jobs by filtering (same params as GET /jobs except for ordering and paging)

I think we could just merge both into a new POST /queue/requeue-job/ by taking the params from 2 and adding in an optional job_ids list, which would act as an additional filter. The other option is to make two different API calls for 1 and 2. I think I prefer the first option, but I go back and forth.

The new backend method to support this is QueueManager.requeue_jobs(). It takes job IDs as input, so a call to JobManager.get_jobs() is needed first for jobs with the normal filter params. The deprecated QueueManager.requeue_existing_job() can be removed once the old REST API view is removed.

PR #45, opened 2016-03-10
Requeue multiple jobs (backend)

dfaller:
This is a pull request to merge the backend changes for issue #16


scale/job/models.py
```diff
@@ -212,55 +206,145 @@ def get_job_updates(self, started=None, ended=None, status=None, job_type_ids=No
...
+        self.filter(id__in=job_ids).update(max_tries=models.F('num_exes') + 1, last_modified=modified)
+
+    def lock_jobs(self, job_ids):
+        """Obtains model locks on the job models with the given IDs (in ID order to prevent deadlocks)
+
+        :param job_ids: The IDs of the jobs to lock
+        :type job_ids: [int]
+        """
+
+        list(self.select_for_update().filter(id__in=job_ids).order_by('id').iterator())
```
wardcr: Could this return a query set? When it gets called by other methods like get_locked_jobs() it could then be chained together into a single query, but once it gets wrapped as a list it forces Django to evaluate the query.

scale/job/models.py
```diff
@@ -435,7 +511,7 @@ class Job(models.Model):
     started = models.DateTimeField(blank=True, null=True)
     ended = models.DateTimeField(blank=True, null=True)
     last_status_change = models.DateTimeField(blank=True, null=True)
-    last_modified = models.DateTimeField(auto_now=True)
+    last_modified = models.DateTimeField(auto_now=True, db_index=True)
```
dfaller: Good catch. Pushed now. I wish PyCharm would auto-add them.

scale/job/models.py
dfaller: Add some documentation to the lock_jobs() method.

scale/recipe/models.py
```diff
@@ -212,81 +287,64 @@ class Recipe(models.Model):
...
-
-        return recipe_jobs
+        :param recipe_ids: The recipe IDs
+        :type recipe_ids: [int]
+        :returns: Dict where each recipe ID maps to a tuple of its recipe model and a list of its recipe_job models
+        :rtype: {int: (:class:`recipe.models.Recipe`, [:class:`recipe.models.RecipeJob`])}
+        """
+
+        # Call get_my_handlers() and organize them, lock dependent jobs, and call get_my_handlers() and organize them
+        recipes = {}  # {Recipe ID: (Recipe, [Recipe job])}
```
wardcr: This return value based on tuples is a bit confusing to follow with all the 0 and 1 based array references in the caller. I think it would be better to use a named tuple for this type of use case: https://docs.python.org/2/library/collections.html#namedtuple-factory-function-for-tuples-with-named-fields

wardcr:
Review complete. My only other comment is that we might want some more tests to cover some of the new methods in jobs/models.py

scale/recipe/models.py
dfaller: Yeah, this makes sense. I'll add this in next sprint where I am already doing some recipe stuff for superseding.

dfaller:
All of the new job model methods are covered in higher level tests. They either replaced an old job model method, in which case I replaced the old unit test, or they are so simple that a unit test doesn't really make sense (like get_locked_jobs()). I think we're good on that front.

Issue #78, opened 2016-03-15
Recipe Names in Product Paths


dfaller:
Add recipe name and version into the paths used to store products in workspaces.

Consider recipe and job name collisions.

cshamis:
The parser is what 'suggests' the storage path to the workspace.  This doesn't sound like strictly a scale thing, but a how you want your parser to parse thing.  Or am I missing something?

cshamis:
I think we're coming around to the fact that we'll /never/ be able to decide how to do the products directory hierarchy.  
Feed/recipe/job/date/exe might be right for some, wrong for others.  And date?  Ingest time or data date?  Maybe we might want feed/recipe/job/date/lat-lon/ 
We *could* do something like have the job define one or more 'logical path' from a template in job definition and automatically build symlinks in the workspace to the physical location (feed/job/datadate/exe). -OR- we could build a WebDAV on the fly napped to the physical.

dfaller:
This would add a ton of complexity for no benefit. The bottom line is that algorithms and users should not care at all what the file path is. Indeed some storage systems (like S3) are going to have performance characteristics that require letting the storage system have control over all or part of the path. Browsing through the storage like a file system to "find" files is an outdated solution. We may not even be able to do this in S3 with even close to good performance. If something outside of Scale wants to access a large number of files, it should query a REST API that can return the appropriate paths/links for the algorithm to access the files it wants. I think creating a base path with things like recipe and job name is still useful because the storage backend may have some benefit to grouping the products that way.

cshamis:
Well... the algorithm developer shouldn't know I agree with you there, but
as a Scale admin, you might have an opinion to how and where things are
stored in the product archive.  Up to and including LAT/LON and the other
things... using a token expansion system to build the product path is a
trivial search and replace.  Hardly a ton of complexity.

But you raise a point, is it the Scale system's job to publish product
information yes/no?  If no, then this doesn't matter.  If Yes, then what is
the obligation?  In the case of S3, then Scale might not

dfaller:
Scale does "publish" product information through the REST API.

The Scale admin shouldn't care about "where" the file is. It doesn't matter. It's some arbitrary string that needs flexibility to work with different storage systems. If you are trying to "search" for files using the file system, it will be much easier and more accurate to use the REST API. Looking at the file system you can't tell if a product is superseded or even published.

[...]

PR #82, opened 2016-03-16
Topic 78

dfaller:
Pull request for issue #78.

### DIFF
--- a/scale/product/configuration/product_data_file.py
+++ b/scale/product/configuration/product_data_file.py
@@ -1,4 +1,4 @@
-'''Defines the product data file input type contained within job data'''
+"""Defines the product data file input type contained within job data"""
 from __future__ import unicode_literals
 
 import os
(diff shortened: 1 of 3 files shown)


scale/product/configuration/product_data_file.py
```diff
@@ -74,11 +76,17 @@ def _calculate_remote_path(self, job_exe, input_file_ids):
         :type input_file_ids: set of int
         :returns: The remote path for storing the products
         :rtype: str
-        '''
-
+        """
+
+        remote_path = ''
+        recipe = Recipe.objects.get_recipe_for_job(job_exe.job_id)
+        if recipe:
```
wardcr: You mentioned recipe/job name collisions and it isn't clear to me how this is handled? It seems like you would want something like a /recipes prefix and /jobs prefix to avoid this block of code having a conflict, or are you just assuming it will never actually happen?

wardcr:
LGTM, pending the answer to the question about recipe/job conflicts.

scale/product/configuration/product_data_file.py
dfaller: It's not handled in a way. A recipe and job with the same type name will put products in the same top level directory (job type name/recipe type name). This doesn't really break anything, but it could be confusing if someone was browsing the file system. In practice I don't think it would occur very often, if at all. We could do some type of a prefix to distinguish recipes and jobs, but I'm not sure John and Chris would like that. Bigger picture I don't think that "browsing" the file system should be a normal use case. I can ask John and Chris about the prefix directories.

dfaller:
Chris, I added the jobs and recipes directories as you suggested. Can you do one final look over? Thanks!

wardcr:
Yes, I think that addresses the concern.

PR #84, opened 2016-03-16
Removed old queue depth code.

wardcr:
- Removed old models, tables, tests, views, docs, and fixture.
- Added a migration to remove the old tables in production.
- The old fixture will need to be manually removed from production.


dfaller:
LGTM. Yeah, good call on the need for disabling the old trigger rule. It doesn't have to be removed (there will be foreign key constraints), it just has to be set so is_active is False. We're going to need to communicate that when it gets merged into develop and master because it will cause errors for existing Scale deployments. I know Adam likes to take tip of develop at times.

PR #85, opened 2016-03-17
closes #77 - adds docker parameters model elements and passes them to…

### LINKED ISSUE #77: Docker parameters for jobs
Create a JSON field holding Docker params for the job_type, job, and job_exe models. The job_type field will be a configurable set of Docker params that the algorithm owner wants to have passed to their Docker container. When a job is created, the Docker params JSON will be passed from the job_type to the job (and so on from the job to each job_exe). When the job_exe is scheduled, cache the Docker params JSON in the RunningJobExecution tasks, and use the values to set the Docker parameters when the task is converted into a Mesos Docker task.

Look into how to best show the user what the full Docker command is.

Create follow-on issue for updates to job type, job, and job_exe REST APIs.

tclarke:
… the mesos task creation.

* adds job type, job, and job execution model elements
* adds an accessor with basic sanity validation
* passes params to mesos task creation
* adds a test for one of the job models

### DIFF
--- a/scale/job/execution/running/job_exe.py
+++ b/scale/job/execution/running/job_exe.py
@@ -42,7 +42,6 @@ class RunningJobExecution(object):
         """
 
         self._id = job_exe.id
-        self._docker_params = job_exe.docker_params
         self._job_type_id = job_exe.job.job_type_id
         self._node_id = job_exe.node.id
         self._node_hostname = job_exe.node.hostname
--- a/scale/job/execution/running/job_exe.py
+++ b/scale/job/execution/running/job_exe.py
@@ -42,6 +42,7 @@ class RunningJobExecution(object):
         """
 
         self._id = job_exe.id
+        self._docker_params = job_exe.docker_params
         self._job_type_id = job_exe.job.job_type_id
         self._node_id = job_exe.node.id
         self._node_hostname = job_exe.node.hostname
--- a/scale/job/execution/running/tasks/base_task.py
+++ b/scale/job/execution/running/tasks/base_task.py
@@ -38,7 +38,7 @@ class Task(object):
         # These values will vary by different task subclasses
         self._uses_docker = True
         self._docker_image = None
-        self._docker_params = job_exe.get_docker_params()
+        self._docker_params = []
         self._is_docker_privileged = False
         self._command = None
         self._command_arguments = None
@@ -88,10 +88,10 @@ class Task(object):
         """Returns the Docker parameters used to run this task.
 
         :returns: The Docker parameters
-        :rtype: List of 2-tuples
+        :rtype: [[str, str]]
         """
 
-        return self.docker_params
+        return self._docker_params
 
     @property
     def id(self):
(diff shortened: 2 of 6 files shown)


scale/job/execution/running/job_exe.py
```diff
@@ -42,6 +42,7 @@ def __init__(self, job_exe):
         """
 
         self._id = job_exe.id
+        self._docker_params = job_exe.docker_params
```
dfaller: I would recommend taking this out since it is not needed.

scale/job/execution/running/tasks/base_task.py
```diff
@@ -83,6 +84,16 @@ def docker_image(self):
...
 
     @property
+    def docker_params(self):
+        """Returns the Docker parameters used to run this task.
+
+        :returns: The Docker parameters
+        :rtype: List of 2-tuples
+        """
+
+        return self.docker_params
```
dfaller: Should be: return self._docker_params
Wouldn't this cause an infinite loop?

scale/job/execution/running/tasks/base_task.py
```diff
@@ -83,6 +84,16 @@ def docker_image(self):
         return self._docker_image
 
     @property
+    def docker_params(self):
+        """Returns the Docker parameters used to run this task.
+
+        :returns: The Docker parameters
+        :rtype: List of 2-tuples
```
dfaller: [(str, str)] would be consistent with how we've been documenting nested types

scale/job/execution/running/tasks/base_task.py
```diff
@@ -38,6 +38,7 @@ def __init__(self, task_id, job_exe):
         # These values will vary by different task subclasses
         self._uses_docker = True
         self._docker_image = None
+        self._docker_params = job_exe.get_docker_params()
```
dfaller: This should be set to []. These Docker params are meant to only be used by the job task, not the pre or post tasks. So the base class should have an empty list (no Docker params). You can do self._docker_params = job_exe.get_docker_params() in the JobTask constructor so it only applies to JobTasks. This is what is meant by "These values will vary by different task subclasses".

scale/mesos_api/tasks.py
```diff
@@ -95,6 +95,9 @@ def _create_docker_task(task):
     mesos_task = _create_base_task(task)
     mesos_task.container.type = mesos_pb2.ContainerInfo.DOCKER
     mesos_task.container.docker.image = task.docker_image
+    # add parameters
+    for k, v in task.docker_params:
+        mesos_task.container.docker.parameters.add(key=k, value=v)
```
dfaller: Does this work with Mesos 0.21? Just want to make it sure it wasn't a new interface or something.

dfaller:
After this issue is merged, we need an issue for REST changes. It would involve adding the docker_params to the appropriate job_type, job, and job_exe serializers. Also adding docker_params to the appropriate job type editing and import/export APIs.

scale/mesos_api/tasks.py
tclarke: Verified with mesos.interface 0.21.2

dfaller:
LGTM now. This PR appears to merge into your own branch. This needs to get merged into develop.

PR #86, opened 2016-03-17
Issue 51

dfaller:
I had to merge Trevor's job model changes for the Docker params into my branch, so this can't be merged until Trevor's issue is merged first.

Highlights for this issue:
1. New job model fields for handling superseded jobs. After more thought, I changed the fields from when I created the issue. Instead of creating two fields to go each direction, I created one as a OneToOneField and we can use the reverse relationship.

2. Updated JobManager.create_job() to handle a new job that supersedes an old job.

3. Create JobManager.superceded_jobs() that marks superseded jobs as such.


tclarke:
lgtm

PR #92, opened 2016-03-21
Issue 81

dickmancj-mil:
Added date filters to Data Feed / Ingest Records view


ericsvendsen-mil:
Need to add `$scope.loading = true;` before calling `getIngests();` inside `$scope.filterResults();` (line 116) to let the user know that the UI is waiting on an API response.

dickmancj-mil:
$scope.loading = true occurs in getIngests();

PR #96, opened 2016-03-22
Issue 70 Adding inputs/outputs to job details

wardcr:
- Adds "inputs" and "outputs" top-level list attributes to the job details view.
- Each input or output consists of a name, type, and value. When a job has not yet run, the name and types of its inputs and outputs will always be defined, but the value may be null. A value can consist of primitives, a file, or an array of files. A file could be a basic scale file or a product file.
- Deprecates the "input_files" and "products" attributes for future removal.
- The serializer implementation is a bit of a hack because the value attribute can contain different types that must be handled dynamically.


dfaller:
LGTM

PR #126, opened 2016-04-01
build-scale fix

dickmancj-mil:
Made changes to ansible playbook that builds scale Docker images. Changes are to utilize local "scale" image rather than geoint/scale to ensure we always deploy the local copy of code to the vagrant environment.


tclarke:
lgtm

PR #134, opened 2016-04-04
Issue 28 Upgrade Django REST Framework

wardcr:
In general, I made separate commits for each version upgrade from 2.x all the way up to 3.3 and the changes are mostly mechanical, but wide-spread. I think we are still not fully taking advantage of REST framework's views, viewsets, mixins, and filtering, but since most of the views require custom model manager methods it isn't as straightforward as the simple CRUD use cases. Therefore, I left that as another refactoring we can look at in the future, since we have a lot to do right now.

- Fixed a few deprecations on the request where upper-cased fields like DATA were changed to lowercase.

- Changes to serializer/field interface. Methods for going to/from native objects are different and the flow is different. In the old version, validating inputs would also create an instance and then save was a second step, but now validation is one step and instance creation + save is a second step. This was a complication for the importer app, which wanted to create validated instances but not save them.

- Replaced custom paging code with the new built-in paging support.

- Tightened down several options where we were being a bit sloppy, which now causes errors. For example, specifying a list field vs. a single value field, dealing with choice fields or read-only fields, and allowing nulls.

- Changed CharField JSON types to use the new built-in JSONField.

- Moved default renderers to configuration instead of pasting it into every view.

- Fixed deprecated configuration options.


scale/mesos_api/api.py
```diff
@@ -84,9 +84,9 @@ class SlaveInfo(object):
     '''Represents information about a host system.
```
dfaller: Replace ''' with """ in this file

scale/metrics/serializers.py
```diff
@@ -1,7 +1,6 @@
 '''Defines the serializers for metrics'''
```
dfaller: Replace ''' with """ in this file

scale/metrics/views.py
```diff
@@ -3,27 +3,25 @@
...
-                                 MetricsTypeDetailsSerializer, MetricsTypeListSerializer)
+from metrics.serializers import (MetricsPlotSerializer, MetricsPlotMultiSerializer, MetricsTypeDetailsSerializer,
+                                 MetricsTypeSerializer)
 
 logger = logging.getLogger(__name__)
 
 
-class MetricsView(APIView):
+class MetricsView(ListAPIView):
     '''This view is the endpoint for retrieving available types of metrics.'''
```
dfaller: Replace ''' with """ in this file

scale/node/serializers.py
```diff
@@ -9,7 +8,7 @@ class NodeBaseSerializer(ModelIdSerializer):
     '''Converts node model fields to REST output.'''
```
dfaller: Replace ''' with """ in this file

scale/node/serializers_extra.py
```diff
@@ -1,5 +1,4 @@
 '''Defines the serializers for nodes that require extra fields from other applications'''
```
dfaller: Replace ''' with """ in this file

scale/node/views.py
```diff
@@ -5,25 +5,25 @@
...
+from node.serializers import NodeSerializer
+from node.serializers_extra import NodeDetailsSerializer, NodeStatusSerializer
 from scheduler.models import Scheduler
 
 logger = logging.getLogger(__name__)
 
 
-class NodesView(APIView):
+class NodesView(ListAPIView):
     '''This view is the endpoint for viewing a list of nodes with metadata'''
```
dfaller: Replace ''' with """ in this file

scale/port/schema.py
```diff
@@ -17,9 +17,9 @@ def __init__(self, key, details):
         '''Constructor sets basic attributes.
 
         :param key: A unique identifier clients can use to recognize the warning.
-        :type key: str
+        :type key: string
```
dfaller: Is this how we should be specifying string types?

scale/port/views.py
```diff
@@ -150,14 +145,12 @@ def post(self, request, *args, **kwargs):
...
+        return Response({'warnings': results})
 
 
 class ConfigurationValidationView(APIView):
     '''This view is the endpoint for validation an exported job and recipe configuration.'''
 
-    renderer_classes = (JSONRenderer, BrowsableAPIRenderer)
-
     def post(self, request):
         '''Validates job and recipe configuration to make sure it can be imported.
```
dfaller: Replace ''' with """ in this file

scale/product/serializers.py
```diff
@@ -27,21 +26,13 @@ class ProductFileSerializer(ProductFileBaseSerializer):
     job_type = JobTypeBaseSerializer()
 
 
-class ProductFileListSerializer(pagination.PaginationSerializer):
-    '''Converts a list of product file models to paginated REST output'''
```
dfaller: Replace ''' with """ in this file

scale/product/views.py
```diff
@@ -3,24 +3,21 @@
...
+from product.serializers import ProductFileSerializer, ProductFileUpdateSerializer
 
 logger = logging.getLogger(__name__)
 
 
-class ProductsView(APIView):
-    '''This view is the endpoint for retrieving a product by filename
-    '''
-    renderer_classes = (JSONRenderer, BrowsableAPIRenderer)
+class ProductsView(ListAPIView):
```
dfaller: Replace ''' with """ in this file

scale/recipe/views.py
```diff
@@ -282,17 +279,16 @@ def get(self, request):
...
+        serializer = self.get_serializer(page, many=True)
+        return self.get_paginated_response(serializer.data)
 
 
-class RecipeDetailsView(APIView):
-    '''This view is the endpoint for retrieving details of a recipe
-    '''
-    renderer_classes = (JSONRenderer, BrowsableAPIRenderer)
+class RecipeDetailsView(RetrieveAPIView):
+    '''This view is the endpoint for retrieving details of a recipe'''
```
dfaller: Replace ''' with """ in this file

scale/scheduler/views.py
```diff
@@ -14,10 +13,9 @@
 logger = logging.getLogger(__name__)
 
 
-class SchedulerView(APIView):
+class SchedulerView(GenericAPIView):
     '''This view is the endpoint for viewing and modifying the scheduler'''
```
dfaller: Replace ''' with """ in this file

scale/source/serializers.py
```diff
@@ -17,21 +16,13 @@ class SourceFileSerializer(SourceFileBaseSerializer):
     pass
 
 
-class SourceFileListSerializer(pagination.PaginationSerializer):
-    '''Converts a list of source file models to paginated REST output'''
```
dfaller: Replace ''' with """ in this file

scale/trigger/serializers.py
```diff
@@ -21,13 +20,7 @@ class TriggerRuleSerializer(TriggerRuleBaseSerializer):
 
 class TriggerRuleDetailsSerializer(TriggerRuleBaseSerializer):
     '''Converts trigger rule model fields to REST output.'''
-    configuration = serializers.CharField()
-
-
-class TriggerRuleListSerializer(pagination.PaginationSerializer):
-    '''Converts a list of trigger rule models to paginated REST output.'''
```
dfaller: Replace ''' with """ in this file

dfaller:
LGTM except for the files I comented as changing ''' to """.

So how do the list fields vs single values work? I think Jared has had issues with the re-queue REST API where he tried to pass a list of job type IDs.

scale/port/schema.py
wardcr: Well my understanding from the Sphinx docs is that either works, but string seems to be preferred by PyCharm.

wardcr:
Well passing a list of values as inputs should still work the same. When we use a simple query parameter, then you just duplicate the parameter: id=1&id=2 and then the view will get a list of ints rest_util.get_int_list('id') will result in [1, 2]. In the case where we accept a JSON payload as POST, then you would just send it as a normal array: ids: [1, 2].

My PR comment refers to a serializer that sends back a list of values in the response. So before we would have this: files = ScaleFileSerializer(), whereas it should be more explicit like this: files = ScaleFileSerializer(many=True) going forward.

tclarke:
No other issues than the ones already mentioned

PR #139, opened 2016-04-07
Added versioned URLs for REST API.

### LINKED ISSUE #107: REST API URL Versions
Add in REST API URL versions, starting with 3.0.
This requires the djangorestframework update to be completed first.

wardcr:
We decided to implement REST API versions using the URL prefix approach. DRF handles this by requiring a prefix for every URL pattern and declaring a namepsace for reverse lookups. Because every URL pattern must be updated, I created a utility method to automate it. This scheme is described here:
http://www.django-rest-framework.org/api-guide/versioning/#namespaceversioning

Therefore, to start a new version, we can simply add 'v4' to the settings file and all the existing APIs will be cloned to the new parent endpoint. Add that point, new serializer or view behavior can be implemented that inspect the requested version and take action accordingly using the request.version attribute.

- All existing endpoints should continue to work.
- All existing endpoints are now duplicated under the /v3 prefix.
- Once the UI, production search indexers, and specific production jobs are migrated we can remove the old endpoints without a version prefix.

Closes #107

### DIFF
--- a/scale/job/views.py
+++ b/scale/job/views.py
@@ -7,7 +7,7 @@ import django.core.urlresolvers as urlresolvers
 import rest_framework.status as status
 from django.db import transaction
 from django.http.response import Http404
-from rest_framework.generics import GenericAPIView, ListAPIView, RetrieveAPIView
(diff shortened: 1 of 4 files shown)


dfaller:
LGTM

PR #162, opened 2016-04-14
Issue 155 - Non-null Strike field

### LINKED ISSUE #155: Make Strike Non-nullable in Ingest
Make the strike field non-nullable in the ingest model. For the migration, have it use a one-off default value of 1. Ingest models with null strike fields should only occur through Trevor's legacy cli tool, so it shouldn't affect anyone.

dfaller:
This is a simple change to make the strike field non-nullable in the ingest model. This helps a number of other pieces of the code which expects the strike field to be populated.

This closes #155.


tclarke:
lgtm

PR #163, opened 2016-04-14
Issue 124 - Default Algorithm Error

### LINKED ISSUE #124: Ensure the default algorithm error condition isn't system error
Help differential between system and algorithm errors by ensuring algorithms don't default to "unknown" system error. Many quickly integrated algorithms don't return appropriate error codes and push the initial diagnosis to the scale/infrastructure team by reporting as system "unknown" error. Reserving system error for pre, post, etc. helps quickly diagnose problems and get the correct team assigned to fix data or algorithm error conditions. Once an algorithm is running the default would be better as ALGORITHM Unknown error.

dfaller:
This issue involved the creation of a default algorithm error that is reported for all job tasks that fail with a non-zero exit code that has no user-defined error mapped. This closes #124.

1. New error added to fixture
2. Small logic change in error mapping code
3. Added unit test

### DIFF
--- a/scale/job/configuration/interface/error_interface.py
+++ b/scale/job/configuration/interface/error_interface.py
@@ -68,21 +68,20 @@ class ErrorInterface(object):
         return self.definition
 
     def get_error(self, exit_code=None):
-        ''' This method retrieves an error given an exit_code.
+        '''This method retrieves an error given an exit code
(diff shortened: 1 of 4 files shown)


tclarke:
lgtm

PR #164, opened 2016-04-14
Issue 130 - Fix pre and post exit codes

### LINKED ISSUE #130: Pre and post task system errors
The pre and post tasks currently use exit codes that cannot fit into 8 bits. It appears that a reported exit code from Mesos is limited to 8-bits, which prevents our system errors from properly being reported.

To fix this, only use positive exit codes from 1 to 255.
Also create a new system error to distinguish operational database errors and include information in the description that explains the causes for operational database errors (e.g. failed to get connections, database deadlock, etc).

dfaller:
This issue looked at the pre and post task exit codes. This closes #130.

1. Created a new error to differentiate database and operational errors, which is the difference between database programming type issues and connection type issues.

2. Changed task exit codes to fit within 8 bits. Apparently Mesos only reports 8 bits worth of exit code, so the large exit codes were being truncated at the binary level and not matching the expected value in the scheduler.

### DIFF
--- a/scale/error/fixtures/basic_errors.json
+++ b/scale/error/fixtures/basic_errors.json
@@ -5,7 +5,7 @@
         "fields": {
             "name": "unknown",
             "title": "Unknown",
-            "description": "The cause of this failure is unknown. System administrators should investigate this failure to determine the true cause and make sure it is correctly tracked in the future.",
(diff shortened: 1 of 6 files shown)


wardcr:
LGTM.
It seems like we need to add a note to the issue that this will require migrating previously saved error values and/or exit codes?

dfaller:
I can add a note talking about that, but I doubt we will bother to do that in our prod environment.

PR #169, opened 2016-04-15
Issue 167 Job type error metrics

### LINKED ISSUE #167: Add error rates to job type metrics
Add new columns to the job type metrics that represent the 3 error categories: SYSTEM, DATA, ALGORITHM.
We would also like to support those values as percentages, but depending on the implementation that may need to be done as a follow-up issue.

wardcr:
Added 3 new columns for job type metrics:
- error_system_count
- error_data_count
- error_algorithm_count

This issue will require a standard DB migration.

Closes #167


dfaller:
LGTM, I'm going to go ahead and merge.

PR #174, opened 2016-04-21
Issue 158 - Builtin Error Metrics

### LINKED ISSUE #158: Builtin Error Metrics
Add the 'built-in' (ones that Scale loads in initial data) errors to the main metrics page by using a new 'is_builtin' field.

I'm undecided on the name of the 'is_builtin' field. Using 'is_system' would be consistent with Scale-loaded jobs, but I didn't know if that would be confusing with the error category of SYSTEM.

1. Add 'is_builtin' field to the error model, default false
2. Set the field to true in all of the errors loaded in by the initial data fixtures
3. Add the new field to appropriate error model serializers.
4. Add these builtin errors to the main metrics page. I think total count is the only metric that makes sense.

wardcr:
- Added a new "is_builtin" field to the error model.
- Updated all errors created via fixtures to be flagged as built-in.
- Added new field to REST services/docs.
- Created a new errors daily metrics table that counts these error types based on job executions.

NOTE: Although there is a migration to add the new field, it will default all existing records to false. Production systems will need to manually change the value to true where applicable.

Closes #158

### DIFF
--- a/scale/docs/interface/rest/error.rst
+++ b/scale/docs/interface/rest/error.rst
@@ -61,6 +61,8 @@ These services provide access to information about registered errors and error m
 +--------------------+-------------------+-------------------------------------------------------------------------------------+
 | .category          | String            | The category of the error. Choices: [SYSTEM, ALGORITHM, DATA].                      |
 +--------------------+-------------------+-------------------------------------------------------------------------------------+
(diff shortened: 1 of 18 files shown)


dfaller:
The fixtures should automatically update the existing builtin errors when the load_initial_data is run.

LGTM, merging.

PR #175, opened 2016-04-22
Issue 156 - Add Error Information to Jobs API

### LINKED ISSUE #156: Add Error Information to Jobs API
Add the error category, title, and description to the API for filtering on jobs. To start with, we'll simply do the join on the error models (should only affect a page worth of jobs) and see if the performance is acceptable. If so, the category and title can be added to the UI with the description as a tooltip over the title.

wardcr:
- Confirmed that error model fields are already included in the jobs list view.
- Added an error_category filter to jobs view, jobs with executions view, and requeue view.
- Fixed some sphinx warnings.

Closes #156

### DIFF
--- a/scale/docs/Makefile
+++ b/scale/docs/Makefile
@@ -51,7 +51,7 @@ html:
 	@echo "Build finished. The HTML pages are in $(BUILDDIR)/html."
 
 code_docs:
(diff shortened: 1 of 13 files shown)


dfaller:
LGTM, merging.

PR #203, opened 2016-04-28
Issue 99 - Source File Detail API

### LINKED ISSUE #99: Source File Detail API
Add a source file detail API for displaying the ancestry file and associated job/recipe details.
Also include the associated ingest details.
We need to be able to do this by both file ID and filename, so two different endpoints might be needed.

wardcr:
- Added a new endpoint /sources/X/, where X can be either a model id or a file name.
- Response includes the source model details and lists of its associated ingests and products.
- Implementation was a bit awkward to workaround circular dependency issues with higher-level models.

Closes #99


dfaller:
LGTM

PR #206, opened 2016-04-30
Issue 202 - Refactored queue status view to follow REST guidelines.

wardcr:
- Refactored the queue status view to follow REST guidelines.
- The refactor adds the missing icon_code field available, as well as the other base job type fields.
- Left the original response structure in place as well for backward compatibility.
- Also added just the icon_code field to the original structure so that it can be optionally used without the UI needing to refactor immediately if desired.


dfaller:
LGTM. Have Eric go ahead and try to use the new structure since he's making changes anyway. He should now be able to add clicking on a row takes you to the jobs filter page, like the running tab. Please create a new issue to remove the deprecated return structure. Hopefully that can be finished this sprint as well.

Issue #182, opened 2016-04-22
Fix Error Category Colors on Overview


dfaller:
On the overview page, the error category colors can change depending on which error categories are present. This can be very confusing to the user. Please set each error category to a permanent color.

Note: This seems to affect the colors in the 'wheel' around the percentage. The error category labels appear to be correct.

dfaller:
Also please change the colors. The green used is confusing in the context of an error color. It would probably be a good idea to change the orange as well. We want the category colors to have good contrast, but not be confused with the green/yellow/red spectrum that is used to indicate the job type health.

wardcr:
This is a good site for picking colors in data visualizations: http://colorbrewer2.org/

I suggest selecting the red column under "Single hue" for this use case, which will give you 3 differentiated shades of red.

wardcr:
I was thinking about this more and I'm getting the feeling we're trying to cram too much into this UI element. How about this design...

- Assign the 3 shades of red for the error categories.
- Color the grid cell background with the most appropriate shade of red based on the errors that are happening. So either we use highest percentage or highest severity.
- Show the breakdown of the errors with colored boxes in the tool-tips, which also acts as a legend and reinforces the scheme. Then we don't need the separate legend in the upper-right.
- Slide the icons up a bit within the cell and move the running count to the bottom center so that the content isn't overlapping and remove the spinners. The counts already imply that things are running and is a bit less distracting than all the spinners.

I think that simplifies the UI, while hopefully still conveying the same general information.

ericsvendsen-mil:
@wardcr I like the idea of simplifying this, and getting rid of the spinners is definitely a good idea since they only spin in Chrome anyway.

I set things up as you suggested - right now the background color is selected based on the type of error that has the highest count, but like you said that's something we can change based on feedback.  I don't know if I like limiting the background for failed job types to 1 color, since that could give a false impression at a glance regarding what's going on, but I think this is a step in the right direction.

What if we partition the cell background color based on failure types?  Since they're all shades of red now, it might work.  Thoughts?

wardcr:
Yeah that might be okay. I think we'd have to try it before I could form a final opinion. It would be nice to show the extra information as long as it doesn't become too busy and hurt the readability. We might be able to make the text a bit more bold or have a slight drop shadow to help alleviate that.

So perhaps try treating each cell as a vertical meter, where it goes lighter to darker from top to bottom?

ericsvendsen-mil:
[...]

PR #215, opened 2016-05-06
Issue 182 - Fix Error Category Colors on Overview

ericsvendsen-mil:


wardcr:
Okay, after seeing this in person, I think the approach can work. It definitely seems easier to read and clearer compared to the old version. I would suggest a couple tweaks:

- The order of the errors and the gradient should be fixed. So the tool-tip should always show System, Data, Algorithm from left to right (like the health meter) and the gradient should also follow that from bottom to top (since that matches typical lighting used in icons). That should reinforce the legend and make things feel more consistent across the grid cells.

- We might want to consider using a similar gradient effect for running vs. completed jobs for consistency? At a minimum I think we should add running count on the same line as completed to reinforce what the count means that's displayed in the cell. So if we do that it, it would probably look nicer if it was styled the same way the errors are.

ericsvendsen-mil:
Sorted errors into System > Data > Algorithm order, and added running count to the tooltip.  I don't think adding a gradient for running vs. completed is necessary - the tooltip is enough to get that across.

wardcr:
LGTM.

ericsvendsen-mil:
Showed this to JP and he brought up some questions. Can possibly be reviewed during the next sprint meeting.

ericsvendsen-mil:
Job types with running jobs now have a background animation; changed data errors to orange as JP suggested; reordered error categories to reflect their severity.

wardcr:
I think the orange is probably okay.

I still have concerns that the pulsing may make the dashboard harder to interpret when you have so many things changing on the screen at once, compared to the single cell on the data feed page. We can try it for a bit and see how others feel. One other idea I had was maybe using a single spinner somewhere towards the top of the grid cells with the total number of running jobs so that you still get a feel for something happening, but in a form that's a bit less overwhelming.

ericsvendsen-mil:
I'm going to merge this but leave issue #182 open so we can make a decision later based on feedback from others.

PR #225, opened 2016-05-06
Issue 181 - Add Error Category Filtering to Jobs Page

ericsvendsen-mil:


wardcr:
The actual feature seems fine. Since we are adding more columns however, I think we might want to try to reclaim some screen space from the time columns.

If it's easy to do, I suggest we omit milliseconds from any time stamp columns like Created and Last Modified, as that level of granularity isn't really needed in the UI. In the Duration column we should omit any fields that are 0. For example "00d, 00h, 00m, 28s" adds a bunch of visual noise vs. just showing "28s". That should free up some space for the error values.

ericsvendsen-mil:
@wardcr I believe I caught everything from our discussion thread re: this issue, and those changes are now implemented.

wardcr:
This is much improved, nice job.

- There are a few tables in the Load place with duration columns that did not get formatted the same.
- Minor typo: There is a period instead of space in the column label "Job Type.Version" on the Load/Running page.
- It looks like there are no static records on the Data Feed/Ingest Records table, so I can't confirm whether the times there are formatted consistently. It would be good to throw in some static data there if it's easy.

ericsvendsen-mil:
Just have 1 record for Ingest Records at the moment - can mock up some more later.  Would be nice if we could get the strike name back from the ingests call.  As it stands now we'd have to do another lookup to get that value.  Otherwise the other things you mentioned should be good now.

wardcr:
I created issue #230 for the strike name.

The Load place Queued table has "Job Type Version", but the Running table has "Version". It would be good to keep it consistent :). The other tables actually just merge the type name and version into one column, but I don't have an issue with having it separate in some places for sorting/filtering purposes.

Other than that, it's good to merge.

ericsvendsen-mil:
Thanks for creating that other issue.  Fixed column name on queued table; will merge this in.

Issue #218, opened 2016-05-06
Document Job Configuration


dfaller:
Document the job configuration schema, which is used in job_configuration fields in both the job and job execution models.

Also remove the job_environment field from the job execution model. I think documentation on job_environment has already been removed, but check this,

PR #228, opened 2016-05-09
Issue 171

tclarke:
Run all system tasks in docker.


scale/ingest/container.py
```diff
@@ -0,0 +1,9 @@
+"""Defines the fields for handling file systems in the local container volumes for ingest and Strike jobs"""
+from __future__ import unicode_literals
+
+import os
+
+from storage.container import SCALE_ROOT_PATH
+
+
+SCALE_INGEST_MOUNT_PATH = os.path.join(SCALE_ROOT_PATH, 'ingest_mount')
```
wardcr: It seems like these types of declared paths should go inside an app-specific settings.py file.

scale/job/configuration/configuration/job_configuration.py
```diff
@@ -0,0 +1,394 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+CURRENT_VERSION = '1.0'
+MODE_RO = 'ro'
+MODE_RW = 'rw'
+
+
+JOB_CONFIGURATION_SCHEMA = {
```
wardcr: Should this be documented under the architecture section of the docs?

scale/job/configuration/configuration/job_configuration.py
```diff
@@ -0,0 +1,394 @@
...
+
+        :returns: The job task Docker parameters
+        :rtype: [:class:`job.configuration.configuration.job_configuration.DockerParam`]
+        """
+
+        params = []
+        for param_dict in self._configuration['job_task']['docker_params']:
+            param = DockerParam(param_dict['flag'], param_dict['value'])
+            params.append(param)
+        return params
```
wardcr: Minor suggestion, but I usually write these types of blocks as follows when possible:
```python
params = self._configuration['job_task']['docker_params']
return [DockerParam(param_dict['flag'], param_dict['value']) for param_dict in params]
```

scale/job/configuration/configuration/job_configuration.py
dfaller: I created a separate issue for this, #218.

scale/ingest/container.py
dfaller: They are more hard-coded than they are configuration. Also some of the "container" modules include functions. I like having them separated into a file specific to "in-container" stuff.

dfaller:
Notes about this PR:

1. Added a new JSON field called 'configuration' to the job model. It holds information about the workspaces and Docker params required by the job. This information is derived from the job's data when it is provided. A data migration was added to populate the new field for all job models with job data (this could be a very long migration!)

2. Added job and configuration fields to the queue model so that the job and workspace information is known by the scheduler when it schedules queued job executions. A data migration was added to populate the new queue model fields from their already linked job execution.

3. Added configuration field to the job execution model so it knows the workspaces and Docker parameters needed for each task. The old Docker param field and the deprecated pre_task_id, job_task_id, and post_task_id fields were removed from the job execution model.

4. Refactored the broker and workspace system to be simpler and removed dependency on using the host's local file system to create directories for a job's input and output data. The input and output data will now reside within named Docker volumes that are passed from task to task.

5. Added code for job executions to configure volume-related Docker parameters when the executions are scheduled.

6. Added Docker repository field to scheduler model and moved scheduler model creation to system initialization instead of through the fixture - I later removed the new scheduler field and replaced it with a settings.py field that can be set by env var.

7. Instead of mounting in a local_settings.py file, Scale will now use environment vars to pass database connection information to system jobs and tasks. All system jobs and tasks have now had their properties switched to use Docker containers.

scale/scheduler/management/commands/scale_scheduler.py
```diff
@@ -6,6 +8,7 @@
 from django.conf import settings
 from django.core.management.base import BaseCommand
 
+from scale import __version__
```
wardcr: Prefer getting the version via the Django settings API to avoid importing higher project-level code into lower-level apps.

scale/storage/brokers/broker.py
```diff
@@ -137,23 +144,183 @@ def upload_files(self, upload_dir, work_dir, files_to_upload):
...
+        that should be mounted into the task container. If uses_mount is False, this property should be None.
+
+        :returns: The remote file system location to mount, possibly None
+        :rtype: string
+        """
+
+        return self._mount
+
+    @property
+    def uses_mount(self):
```
wardcr: Seems like uses_mount() could be deleted and just have mount() return a value or None only. Since the documentation requires that behavior anyway, it's one less thing to keep in sync.

scale/storage/brokers/host/host_broker.py
```diff
@@ -0,0 +1,99 @@
+"""Defines a broker that mounts a local host directory into the task container as its backend storage"""
```
wardcr: Why is this module in it's own package instead of just in /brokers like the others?

scale/storage/test/test_models.py
```diff
@@ -11,6 +12,7 @@
 import datetime
```
wardcr: Fix import ordering.

scale/job/migrations/0016_auto_20160421_1553.py
```diff
@@ -0,0 +1,65 @@
...
+            configuration = JobConfiguration()
+            for name in input_workspaces:
+                configuration.add_job_task_workspace(name, MODE_RO)
+            if not job.job_type.is_system:
+                for name in input_workspaces:
+                    configuration.add_pre_task_workspace(name, MODE_RO)
+                    # We add input workspaces to post task so it can perform a parse results move if requested by the
+                    # job's results manifest
+                    configuration.add_post_task_workspace(name, MODE_RW)
+                for workspace in Workspace.objects.filter(id__in=data.get_output_workspace_ids()).iterator():
```
wardcr: If we're concerned with how long this migration might take to complete, it might help to hoist the Workspace query outside all the loops. Since the number of workspaces is small, we could just query for all of them up front and cache it.

scale/scheduler/models.py
```diff
@@ -10,27 +12,36 @@
...
         :rtype: :class:`scheduler.models.Scheduler`
-        '''
+        """
         try:
             return Scheduler.objects.get(pk=1)
         except Scheduler.DoesNotExist:
             logger.exception('Initial database import missing master scheduler: 1')
             raise
 
+    def initialize_scheduler(self):
```
wardcr: Based on the PR note about the scheduler fixture being removed in order to add a DB field that was also later removed...should we go back to just having the fixture or is there still a compelling reason to use this custom initialization code?

wardcr:
Review complete.

scale/storage/brokers/host/host_broker.py
dfaller: As we grow the number of brokers, I figured it would be useful to have their own directories for additional files/code they might need. The NFS broker not in its own directory is the old one that will be deleted once we create the new one (in its own directory).

scale/scheduler/models.py
dfaller: The initialization code is nice because it only runs once. When the fixture runs, it overwrites existing model values. So whenever we do a Scale deploy, the scheduler is always automatically unpaused, which is not ideal or expected for a user.

I think we should use "run once" initialization code for database fields that act as configuration that needs to be set up once, but then should honor the user's settings (Scheduler model) and use fixtures for models that need to be updated as Scale is in order to function properly (Job Type models, etc).

scale/storage/brokers/host/host_broker.py
wardcr: I think it's more Pythonic to start with just the single module and then create a sub-package if/when you actually need one so you don't have useless directories. See for example Django views.py or tests.py.

PR #231, opened 2016-05-10
Issue 229 PR - Opening up permissions on output_data directory

dfaller:
I changed the Dockerfile to open up the permissions to 777 on the /scale/output_data so that algorithm job containers can run as any users and have appropriate access to write output files.


tclarke:
LGTM
Will merge slightly differently in dockerfile based build.

PR #235, opened 2016-05-11
Issue 229 - Fix output_directory permissions

dfaller:
I updated Scale so it now mounts the output_data Docker volume into the Scale pre task so that it gets initialized with the correct permissions.


wardcr:
LGTM

tclarke:
LGTM

Issue #190, opened 2016-04-25
Import errors when null max_scheduled or author_url


dfaller:
The import REST API now errors when max_scheduled or author_url is not provided for a job type. This may be a result of the DRF update.

wardcr:
I tried to create a unit test that does a job type import with only the minimum required fields and it worked, so I am currently unable to reproduce the issue.

wardcr:
Okay after further troubleshooting, I found out the root cause. I was testing this case by omitting the optional fields, which does work. However, if you send an optional integer field with a null value or an optional string field with an empty string value, then it gets rejected. This comes up when a user copies/pastes the output of say the job-type details service as a starting point for the import because that's how those fields get rendered when they are not filled in. We can see if there is an easy way to relax the validation so that these cases do not produce an error, but in the short term it's easy to workaround by just deleting the fields from the input.

PR #237, opened 2016-05-12
Issue 190 - Import errors when null max_scheduled or author_url

wardcr:
- Fixed import serializers to accept blank or null values for optional fields.
- Added a migration for recipe type to make the fields consistent with the other models.


dfaller:
LGTM

wardcr:
Closes #190

wardcr:
Closes #190

Issue #187, opened 2016-04-22
Re-position Error Description on Job Detail Page


dfaller:
Some of the error descriptions have increased in length and we need to change how they are displayed on the job detail page so they look nice.

ericsvendsen-mil:
Closed by #247

PR #247, opened 2016-05-12
Issue #187: Re-position Error Description on Job Detail Page

ericsvendsen-mil:
The first two entries in the jobs grid are failed jobs. The first has a long error message and the second has a short one.

### DIFF
--- a/scale-ui/app/modules/jobs/partials/jobDetailTemplate.html
+++ b/scale-ui/app/modules/jobs/partials/jobDetailTemplate.html
@@ -17,10 +17,13 @@
 
 <div class="tab-content">
     <div class="tab-pane active" id="details">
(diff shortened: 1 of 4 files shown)


wardcr:
The error message text is a little cumbersome to read on a widescreen monitor because it spans the entire width of the page. I think we should put a max-width on it.

I'm also not sure that we need both the error badge and the error message box? It takes up a bit more room and we already have the same information below in the table. If we want to keep the badge still, we could maybe put them on the same line to condense a bit too?

Lastly, I noticed the gray border at the top of the tab panel stops abruptly at around 80% of the width for some reason for me.

ericsvendsen-mil:
Reformatted error description and fixed top border issue.

wardcr:
LGTM

PR #256, opened 2016-05-16
Issue 217 - New NFS broker

dfaller:
1. Created an NFS broker that works with the "new" fully containerized system. Broker now has a new "volume" field that replaces the old "mount" field. The volume gives information to the scheduler about Docker volumes that must for mounted for the workspaces used by a job execution.

2. Removed the old NFS broker and its base class.

3. Updated the workspace/broker documentation.

4. Added broker unit tests.


scale/job/test/test_models.py
```diff
@@ -18,11 +18,11 @@
 from job.configuration.data.job_data import JobData
 from job.configuration.interface.error_interface import ErrorInterface
 from job.configuration.interface.job_interface import JobInterface
-from job.execution.container import get_job_exe_input_vol_name, get_job_exe_output_vol_name, SCALE_JOB_EXE_INPUT_PATH, \
-    SCALE_JOB_EXE_OUTPUT_PATH
+from job.execution.container import get_job_exe_input_vol_name, get_job_exe_output_vol_name, \
+    get_workspace_volume_name, SCALE_JOB_EXE_INPUT_PATH, SCALE_JOB_EXE_OUTPUT_PATH
```
wardcr: Instead of importing all these individual functions, I think it's clearer to import the container namespace. You can still do a separate from import for the constants if you like. Also would prefer () line continuation instead of \.

scale/job/test/test_models.py
```diff
@@ -297,32 +297,47 @@ def test_get_latest_job_exes_with_a_filter(self):
         self.assertDictEqual(latest_job_exes, expected_result, 'latest job executions do not match expected results')
 
     def test_schedule_job_executions(self):
```
wardcr: This test case has become pretty complex. I think it's good to have 1 that mixes different uses together, but it would be nice to have some very simple ones too that are easier to follow. Cases like using no volumes or just using a RO for inputs and no outputs with a single workspace and job type.

scale/storage/brokers/nfs_broker.py
```diff
@@ -5,148 +5,95 @@
 import os
 import shutil
 
-from storage.brokers.broker import OldBroker
-from storage.nfs import nfs_umount, nfs_mount
```
wardcr: Is the storage.nfs module still needed after this refactor?

scale/storage/test/brokers/test_nfs_broker.py
```diff
@@ -1,391 +1,203 @@
...
+        file_1 = ScaleFile()
+        file_1.file_path = workspace_path_file_1
+        file_2 = ScaleFile()
+        file_2.file_path = workspace_path_file_2
+        file_1_up = FileUpload(file_1, local_path_file_1)
+        file_2_up = FileUpload(file_2, local_path_file_2)
 
         # Call method to test
-        mountstats_data = '''16 36 0:3 / /proc rw,nosuid,nodev,noexec,relatime shared:5 - proc proc rw
+        mountstats_data = """16 36 0:3 / /proc rw,nosuid,nodev,noexec,relatime shared:5 - proc proc rw
```
wardcr: Is this big string dump really needed for the test?

wardcr:
LGTM, pending comments.

scale/storage/brokers/nfs_broker.py
dfaller: Yes, it is still used by ingest and strike. Once the ingest system has been refactored similar to this broker refactor, the nfs module can go away.

scale/storage/test/brokers/test_nfs_broker.py
dfaller: I'm not sure. This was originally Trevor's code. I think bbcp uses it to see the NFS mount locations. I'm thinking we'll probably have to rework the bbcp stuff anyway now that everything is in a Docker container.

PR #261, opened 2016-05-18
Issue 241 - REST API for editing workspaces

wardcr:
- Added model method, view, tests, and docs for editing a workspace.
- Fixed a bug where error model edits were not actually being saved that was discovered while looking up example code to use for workspace edits.


tclarke:
lgtm

PR #269, opened 2016-05-19
Issue 246 - Update Strike API to follow REST standards

wardcr:
- Added endpoint /strikes/ that supports list and create.
- Added endpoint /strikes/ID/ that supports details and update.
- Added endpoint /strikes/validation/ for validating configuration.
- Improved validation/error messages to match other APIs and address the original reported issue.
- This patch also removes the old /strike/create/ endpoint.

- Fixed bugs with the recent workspace changes, mainly typos and a couple things where the implementation didn't match the docs.


dfaller:
LGTM

wardcr:
- Updated scale-cli to use the new create endpoint.
- Removed old REST view since the UI currently does not use it.

Issue #184, opened 2016-04-22
New page for job type error metrics


dfaller:
Create a new page to display job type error metrics. You should be able to use the metrics API to get back the required data.

ericsvendsen-mil:
Reopening the issue since it's not yet resolved.  Still working to add failure rates, total job counts, changing time window, and ordering the grid by highest failure percentage descending.

ericsvendsen-mil:
Adding comments from @wardcr here for posterity:

* Having color on every single value I think will make it a bit hard to recognize important trends. I think we should consider some sort of scheme that introduces more variation. For example, only color the values with the highest count within a table cell or turn it into more of a progress bar where the width of the color is proportional to the % of the errors.
* Add a tool-tip over each value that shows the value as a % and the error category label.

PR #270, opened 2016-05-19
Issue #184 - New page for job type error metrics

ericsvendsen-mil:
This needs to be tested against a live API.


wardcr:
This is on the right track of the print out I saw, here are some suggestions.

- [ ] Change the colors to the new scheme from issue #182  (you already mentioned it)
- [ ] I'm not too keen on the name "Performance", as that implies to me how fast it runs or much it uses in terms of resources. I think something like "Error Rates" would be better. Also move this tab to the last position after Jobs because it makes a bit more logical flow left to right.
- [ ] I'm seeing an extra vertical scroll bar at the page level on my screen.
- [ ] The date columns should be more like 24 hours, 48 hours, 72 hours, 1 week rather than consecutive days. I realize this requires multiple queries to the backend for the moment. Also the time should be newest to oldest from left to right to emphasize more recent changes.
- [ ] Add a total error count at the beginning of each table cell breakdown.
- [ ] Having color on every single value I think will make it a bit hard to recognize important trends. I think we should consider some sort of scheme that introduces more variation. For example, only color the values with the highest count within a table cell or turn it into more of a progress bar where the width of the color is proportional to the % of the errors.
- [ ] Add a tool-tip over each value that shows the value as a % and the error category label.

ericsvendsen-mil:
I fixed a few of the basic things you mentioned, but I think in the spirit of iteration we should merge this in and then move on from there.

dfaller:
I'm going to merge this in to finish up the sprint. We'll gather more feedback and do additional changes.

PR #286, opened 2016-05-20
Issue 230 - Updated REST docs for ingest views.

wardcr:
This issue was requested when the ingest records table was updated in the UI. After reviewing the code more closely, it seems this was actually already implemented. So I just updated the documentation to make it more clear and fixed some warnings.


dfaller:
LGTM, I'm going to go ahead and merge to get this in for this sprint.

PR #291, opened 2016-05-24
Issue #184 - New page for job type error metrics

ericsvendsen-mil:
Added fail rate percentage; adjusted time window; added dynamic opacity to error labels based on each error's percentage of total jobs; ordered by failure rate descending for a 24 hour period; other misc changes

### DIFF
--- a/scale-ui/app/modules/jobs/controllers/jobTypesErrorRatesController.js
+++ b/scale-ui/app/modules/jobs/controllers/jobTypesErrorRatesController.js
@@ -51,6 +51,7 @@
                 displayName: '24 Hours',
                 enableSorting: false,
                 enableFiltering: false,
+                headerCellTemplate: 'gridHeader.html',
                 cellTemplate: 'gridRow.html'
             },
             {
@@ -58,6 +59,7 @@
                 displayName: '48 Hours',
                 enableSorting: false,
                 enableFiltering: false,
+                headerCellTemplate: 'gridHeader.html',
                 cellTemplate: 'gridRow.html'
             },
             {
@@ -65,6 +67,7 @@
                 displayName: '30 Days',
                 enableSorting: false,
                 enableFiltering: false,
+                headerCellTemplate: 'gridHeader.html',
                 cellTemplate: 'gridRow.html'
             }
         ];
@@ -84,6 +87,20 @@
                     });
                 }
             });
+        };
+        
+        vm.setStyle = function (colField, errorType) {
+            var errorValue = (colField[errorType] / colField.total).toFixed(2);
+            var textColor = errorValue >= 0.5 ? '#fff' : '#000';
+            var rgb = errorType === 'system' ? '103, 0, 13' : errorType === 'algorithm' ? '203, 24, 29' : '241, 105, 19';
+            return 'background: rgba(' + rgb + ', ' + errorValue + '); color: ' + textColor;
+        };
+        
+        vm.getPercentageOfTotal = function (errorTotal, total) {
+            if (total === 0) {
+                return '100%';
+            }
+            return ((errorTotal / total) * 100).toFixed(0) + '%';
         };
 
         vm.filterResults = function () {
@@ -169,7 +186,7 @@
                             if (d.twentyfour_hours.total > 0) {
                                 return d.twentyfour_hours.errorTotal / d.twentyfour_hours.total;
                             }
-                            return 0;
+                            return 1;
                         }, ['desc']);
 
                         vm.gridOptions.data = vm.performanceData;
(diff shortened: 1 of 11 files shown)


scale-ui/app/modules/jobs/controllers/jobTypesErrorRatesController.js
```diff
@@ -64,6 +88,20 @@
                 }
             });
         };
+        
+        vm.setStyle = function (colField, errorType) {
+            var errorValue = (colField[errorType] / colField.total).toFixed(2);
+            var textColor = errorValue >= 0.5 ? '#fff' : '#000';
+            var rgb = errorType === 'system' ? '103, 0, 13' : errorType === 'algorithm' ? '203, 24, 29' : '241, 105, 19';
```
tclarke: should this be set in css and use classes instead of hard coding the color values?

scale-ui/app/modules/jobs/partials/jobTypesErrorRatesTemplate.html
```diff
@@ -1,12 +1,53 @@
-<ais-header hide-title="true" name="'Job Type Error Rates'" show-subnav="true" subnav-links="vm.subnavLinks" loading="vm.loading"></ais-header>
-<div class="page-header">
-    <h1>
-        Job Type Error Rates
-        <small><span class="label label-system">System Errors</span> <span class="label label-data">Data Errors</span> <span class="label label-algorithm">Algorithm Errors</span></small>
-    </h1>
-    <hr />
-</div>
+<ais-header name="'Job Type Error Rates'" show-subnav="true" subnav-links="vm.subnavLinks" loading="vm.loading"></ais-header>
+<!--<div class="page-header">-->
```
tclarke: Remove commented code

scale-ui/app/modules/jobs/controllers/jobTypesErrorRatesController.js
ericsvendsen-mil: Ideally yes, but the alpha value ("a") in rgba needs to be dynamic and you can't set that individually apart from the rgb values.  Setting the opacity on the element would affect the background and the text, which is not desirable.  Therefore it needs to be done like this.

ericsvendsen-mil: Done.

tclarke:
LGTM

wardcr:
This is much improved over the last version and pretty close to the original print out now, nice work.

- In the last review I suggested "Error Rates" as the title, I think we can omit the Job Type prefix since you're already in the job section.
- Use a consistent label, so either "Error" or "Failure". Currently the page header and table header are different.
- I'm not sure the failure rate value should be represented with the color blue, as I think it causes a bit of confusion. I suggest making the table header text "Failure Rate" just plain black or gray and then using a shade of red for the progress meter. We could possibly use the shade of red corresponding to the error type with highest count.
- Similarly I don't think we should use green for "total jobs", since that implies success. You could use the blue here instead.
- Since this page is focused on errors, we might want to put the Total values after the error values.

ericsvendsen-mil:
Thanks @wardcr.  I implemented your suggestions - decided to go with "Failure" since that's what is used on the printout.  I made the failure rate bar the color of the highest percentage error, but I think that might get confusing - I don't want people thinking that overall failure percentage applies to any one error type.  I would prefer using gray for that bar, but I included the colors so you can see what it looks like.

wardcr:
Yes, I could understand it possibly causing confusion. My thought was that since the breakdown is shown right next to it, that would be sufficient to avoid the issue. Having said that, I'm not opposed to using a single color. Maybe get feedback from the stakeholders? Gray might be a problem if it clashes with the zebra striping.

- If we are going with "Failures", then I suggest changing the URL route to match too.
- I noticed if I make the window a bit smaller, the overall failure rate % is word-wrapping and hidden. It should probably be forced no-wrap and potentially have the bar shrink more as space gets smaller.
- It looks like there is a tiny extra table column on the right side for no reason.

ericsvendsen-mil:
Changed route, made failure rate bars gray, and fixed wrapping issue.

The extra table column you see is something I've been trying to fix for quite a while with all the grids in the app.  The problem is that it seems to come and go so nailing down how to fix it has been quite troublesome.  For example, I've seen it before on Chrome on my machine, but I don't see it now.  That's something I'll continue to look into, but it shouldn't hold up this PR.

wardcr:
Okay, that's fine.

One other thought I had...what do you think about showing % values inside the colored error cells instead of raw values? My reasoning is that when skimming over the values it's very hard to mentally grasp the proportions between the values when you have a bunch of 3 digit numbers, whereas a % is much faster to grok. We could still show the raw values in the tool-tip, like you currently have % there. And we could leave the total as a raw value so that you could quickly grasp the scale of the % values as well. It might deviate from the printout, but I feel like it might be more usable?

ericsvendsen-mil:
I honestly don't know.  I could see that being a personal preference sort of thing.  Perhaps it's best to present the percentage/raw values options to a user base and see what is preferred?

Issue #232, opened 2016-05-10
Error Color and Text of Master and Scheduler Status on Overview


dyerz:
If the Mesos master or scale scheduler cannot be reached, the default color for the status boxes is green. This may confuse administrators who are not familiar with the UI. It would be great if these boxes could respond like the Heath, Jobs, and Job Load sections, where the color of the boxes is red and an error message is displayed.

PR #292, opened 2016-05-25
Issue #232 - Error Color and Text of Master and Scheduler Status on Overview

ericsvendsen-mil:
I decided to go with a gray status color when the cluster status is unreachable since that conveys that the status is unknown.  Screenshot attached. 
image

### DIFF
--- a/scale-ui/app/modules/common/factories/gaugeFactory.js
+++ b/scale-ui/app/modules/common/factories/gaugeFactory.js
@@ -28,6 +28,7 @@
                 this.config.greenColor 	= configuration.greenColor || '#8fca0e';
                 this.config.yellowColor = configuration.yellowColor || '#ffc317';
                 this.config.redColor 	= configuration.redColor || '#f54d36';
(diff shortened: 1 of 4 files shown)


wardcr:
I'm okay with the UI appearance, but I'm curious if there are actual concrete cases where the cluster is unreachable and that *isn't* a problem? If there is, then this is fine, but if not then it should probably be represented as an error from an admin's point of view.

ericsvendsen-mil:
From my point of view, an unreachable cluster could be a result of localized network problems or other things completely unrelated to the system itself, therefore there's no way of knowing whether an actual error occurred or not.  I was just trying to differentiate when an actual error happened that can be reported vs. an unknown.

Whatever makes the most sense is fine with me though - my understanding of this case is admittedly vague.

wardcr:
I would suggest then if you aren't aware of something specific, from an admin's point of view any type of network problem is still something they want to investigate and draw attention to. You can still leave the sub-text as is to indicate that the machine is unreachable rather than a more specific error.

ericsvendsen-mil:
image

Issue #212, opened 2016-05-04
Job Table Sorting Issue


dyerz:
The previously selected sort order for the job table does not save correctly.

Steps to reproduce:

1. Go to the jobs table
2. Select the drop-down on the Created table and choose Sort Descending
3. Navigate away from the jobs table
4. Navigate back to the jobs table
5. The Created column still shows the icon indicating it is sorted descending however the data is actually sorted ascending

PR #297, opened 2016-05-26
Issue #212 - Job Table Sorting Issue

ericsvendsen-mil:
Fixed state management for sorting and filtering the jobs table.  Also updated the way the admin user object is handled, so we may want to double check that all the "admin" features still show/hide properly.


wardcr:
LGTM

PR #299, opened 2016-05-27
Fix mesos/docker force pull - Issue 294

### LINKED ISSUE #294: Docker images are not force pulled on hosts
When an algorithm image is cached on a slave node and is updated, it's not force pulled. I believe :latest will force pull and other tags won't unless the force_pull_image option is setup in the DockerInfo for the mesos task.

tclarke:
closes #294


wardcr:
The actual fix LGTM. Is the removing of mesos.py and all the changes under mesos just dead code from an earlier refactor or upgrade?

wardcr:
Sorry, just saw your commit comments...so do these files get handled when the image is built now instead?

tclarke:
These were used by the disconnected test setup on Windows. The files are from an earlier version of mesos and aren't compatible with our current setup. These are part of the mesos.interface module.

Issue #259, opened 2016-05-17
Table State Issue


dyerz:
This may relate to issue #212 

There is an issue where the Jobs Table shows the columns from the Ingests Record table.

1. Go to Ingests Record table
2. Sort by file name
3. Go to Jobs table

The columns from the ingest table will be displayed.

PR #300, opened 2016-05-27
Issue #259 - Table State Issue

ericsvendsen-mil:
Updated recipes and ingest records grids to use stateService.  Load and running grids currently do not have any filtering/sorting capabilities so those were left alone for now.


wardcr:
LGTM

Issue #296, opened 2016-05-26
Job/Recipe Type Sorting


dickmanp:
When viewing the list of job or recipe types (or selecting a filter for them) the list appears to be sorted alphabetically.  However some of the names are not in the appropriate order.  I am assuming that we are sorting on one thing (like name?) and displaying another (like title?).

PR #301, opened 2016-05-31
Issue #296 - Job/Recipe Type Sorting

ericsvendsen-mil:


wardcr:
- It looks like this PR is sorting by name+version, however the change should be to sort by title+version. The problem is that the system name field does not always match the title display field, which causes the results to appear out of order to the user since title is the value being shown.

- It looks like the sorting is being done client-side, but the corresponding services support server-side sorting. Will this work when the user selects other sort columns and the results are paged?

- It looks like this PR only addresses jobs tables, but I believe the same issue exists for recipes as well?

ericsvendsen-mil:
@wardcr all good points.  Working these out now...

wardcr:
Okay, I was asking about the recipes because of the addition to the stateService for job types...I was expecting to see a similar block for recipes.

ericsvendsen-mil:
I see now.  That is due to poor naming conventions on my part.  The stateService modifications were to bring the jobTypesFailureRates grid up to speed with the other grids.  I should probably name that object accordingly instead of just generically calling it "jobTypesParams".

ericsvendsen-mil:
The latest commit should address the issues mentioned by @wardcr.  The sorting of job types and recipe types and datagrid sorting won't conflict due to the fact that sorting for these 2 data types in those grids only relates to the order in which they appear in the filter dropdown.  It doesn't affect the sort order of the jobs and recipes themselves - that is still controlled by the datagrid column controls.

PR #303, opened 2016-06-01
Fixed #289 by changing exit code parse regexp to work with older and …

### LINKED ISSUE #289: Fix Exit Code Parsing
In the newer versions of Mesos (both 0.25 and 0.24 in DevNet), Scale is no longer parsing exit codes from Mesos. This results in no exit code information for Scale and all errors end up being unknown.

Two possibilities for the bug are:
1. Using the incorrect version of the Mesos Python egg
2. Mesos updated their interface (we are parsing the exit code from the "message" string field of a Mesos TaskStatus, so it is fragile)

tclarke:
…newer message format.
closes #289

### DIFF
--- a/scale/mesos_api/utils.py
+++ b/scale/mesos_api/utils.py
@@ -12,7 +12,7 @@ except ImportError:
 
 
 EPOCH = datetime.utcfromtimestamp(0).replace(tzinfo=utc)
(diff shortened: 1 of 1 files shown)


wardcr:
LGTM

PR #305, opened 2016-06-02
Initial merging of dcos and standard scale images

tclarke:


dfaller:
I would like to have Dan take a look at this.

droessne:
Almost everything looks good except the changing of the UI in the scale.conf file. this is needed for proper dcos proxy redirection. Also why are you removing the DCOS scale DB directory? This is needed if the user is not using a pre-built database.

tclarke:
@droessne It has to be rooted at /ui/ ? Can't we handle that in the haproxy config so apache see's / and haproxy rewrites from /ui/ to /? The ui is going to incorrect for most other deployments so our other option is to template the conf file or possibly use a variable (I'd have to look into apache variables to make sure they can be used in a Location)

As for the db, are you referring to the scale db image? It should be using the mdillon/postgis image from the docker hub now instead of requiring a custom image.

droessne:
The proxy is not configurable it is built into DCOS.

The DB is fine then i just wanted to make sure we had something.

tclarke:
ok, I'll make a change to fix the ui prefix issue

dickmancj-mil:
LGTM

Issue #316, opened 2016-06-06
Accommodate larger amount of job types on Overview page


ericsvendsen-mil:
If there are a lot (100+) of job types in the system, the Jobs grid on the Overview page doesn't show status for all of them, plus the user is unable to scroll to see all the cells.

PR #317, opened 2016-06-06
Issue #316 - Accommodate larger amount of job types on Overview page

ericsvendsen-mil:
Increased the page size for now to address the job status part of issue #316.  This will only work until there are >1000 job types, so more work needs to be done in order to arrive at a more permanent solution.  For now though this should be merged in to fix a bug that's hindering users on-site.

### DIFF
--- a/scale-ui/app/modules/jobs/directives/jobHealthDirective.js
+++ b/scale-ui/app/modules/jobs/directives/jobHealthDirective.js
@@ -8,7 +8,7 @@
         $scope.jobHealth = {};
 
         var getJobTypeStatus = function () {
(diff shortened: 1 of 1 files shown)


wardcr:
LGTM. You might also consider sorting them by say number of running and then number of completed/failed. That way active content is more likely to fit into the view without scrolling/paging compared to maybe a bunch of jobs that haven't run in a while.

ericsvendsen-mil:
Yes, absolutely.  Paul, Derick, and I were talking about this earlier, and that is part of the longer-term strategy I'll be working on next.

PR #319, opened 2016-06-06
Issue #318 - Job type failure rates page encounters 500 error

ericsvendsen-mil:


wardcr:
LGTM. Just to double check...are there any corresponding data files that need to be updated as well for static data mode?

ericsvendsen-mil:
No, the random data generator piece in `app/test/scripts/backendStubs.js` doesn't validate any columns it receives.  It just creates data for each value in the columns array and sends it back.

PR #327, opened 2016-06-08
Issue 326 - Restrict configuration import upload view

wardcr:
- Removed default parsers so only a multi-part form post is accepted.
- Added a custom renderer so the default HTML form is not displayed.
- Added a more detailed description to explain the purpose of the view that is displayed in the browseable API.


dfaller:
LGTM

PR #333, opened 2016-06-10
Issue 180 - Docker image updates

droessne:
Trevor, i have added the sudo's and removed the PREFIX. running it off of root appears to work. Also i figured out the problem we had yesterday when we left. it was due to the scale user (7498) not being present on the slaves. it appear mesos was trying to chown something to "scale" and we didn't have a scale user on devnet. this could be an issue in the future with people testing out scale in their environments.


tclarke:
LGTM

PR #342, opened 2016-06-16
Issue 52 - Recipe Superseding

dfaller:
This is a big one. It puts in place the backend support for recipes superseding one another.

Highlights:
Updated recipe model with new fields related to superseding
Updated recipe_job model to support jobs being in multiple recipes (as they supersede one another)
Data migration requires doing an entire re-write of the recipe_job table due to primary key change
Created a RecipeGraphDelta class that figures out the difference between two recipe definitions
Updated the logic for creating and handling recipes to support recipes superseding each other (as well as their jobs)


tclarke:
LGTM

PR #343, opened 2016-06-16
Issue #340 - Job Detail File Link Bug

### LINKED ISSUE #340: Job Detail File Link Bug
On the job detail page (and anywhere else that displays files), no link should be generated when the file's URL field is null. If the URL field is null, just display the file name as standard text with no anchor tag. This should apply to both source files and products.

ericsvendsen-mil:
Fixes #340


dfaller:
LGTM

PR #346, opened 2016-06-17
Issue #341 - Metrics Unit Conversion

### LINKED ISSUE #341: Metrics Unit Conversion
The metrics page needs to perform useful unit conversion depending on the "units" value:

"bytes" should be converted into a human readable size so that the value is between 0 and 1024
"seconds" should be converted into minutes, hours, days, etc. in a way that is human readable

The converted, human readable values should be used on both the graph axis and the popup display values when you mouse over.

ericsvendsen-mil:
resolves #341


dfaller:
LGTM

dfaller:
I'm going to go ahead and merge this.

PR #348, opened 2016-06-17
minor updates

ericsvendsen-mil:
corrected workspace broker fields; bugfix for node page height

### DIFF
--- a/scale-ui/app/modules/charts/grid/gridChartController.js
+++ b/scale-ui/app/modules/charts/grid/gridChartController.js
@@ -56,6 +56,9 @@
                 return;
             }
             height = (Math.ceil(newValue.length / (Math.floor(width / $scope.cellWidth))) * $scope.cellHeight);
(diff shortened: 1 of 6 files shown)


tclarke:
LGTM pending travis checks

dfaller:
It looks like you replaced "nfs_path" with "mount" in the nfs broker. The field really is "nfs_path".

"host_path" is correct for the host broker.

ericsvendsen-mil:
I changed these based off what was being returned by the API this morning.  After testing on-site everything worked.  I can change "mount" back to "nfs_path" if that is indeed what the api will come back with though.

dfaller:
nfs_path is indeed correct for the nfs broker.

I think what you saw was an outdated config in a legacy workspace in dev. I will correct it to prevent future confusion.

ericsvendsen-mil:
Makes sense, thanks for catching that.

dfaller:
LGTM now

PR #356, opened 2016-06-22
Issue 53 - Product Superseding

dfaller:
Updates to how products are handled for superseded jobs/recipes:

- New is_superseded and superseded fields added to the product_file model
- Superseded jobs do not publish products
- When a job publishes its products, it also unpublishes the products of any job(s) it supersedes
- When a new product is published, any previously published product with the same UUID is superseded and unpublished
- When a recipe job is superseded with no new job to "replace it", its products are immediately unpublished


scale/product/test/utils.py
```diff
@@ -28,8 +28,9 @@ def create_file_link(ancestor=None, descendant=None, job=None, job_exe=None, rec
                                            recipe=recipe)
 
 
-def create_product(job_exe=None, workspace=None, has_been_published=False, file_name='my_test_file.txt',
-                   file_path='/file/path/my_test_file.txt', media_type='text/plain', file_size=100, countries=None):
+def create_product(job_exe=None, workspace=None, has_been_published=False, is_published=False, uuid='',
```
wardcr: I think it would be better for debugging to default uuid=None and then set a more realistic value:
if not uuid:
    builder = hashlib.md5()
    builder.update(job_exe.job.job_type.id, file_name)
    uuid = builder.hashdigest()

scale/recipe/models.py
```diff
@@ -157,11 +157,18 @@ def _create_recipe_jobs(self, recipe, event, when, delta, superseded_jobs):
             recipe_job.recipe = recipe
             recipe_jobs_to_create.append(recipe_job)
 
-        # Supersede any jobs that were changed or deleted in new recipe
         if delta:
-            for superseded_job_name in delta.get_deleted_nodes():
-                jobs_to_supersede.append(superseded_jobs[superseded_job_name])
+            # Go through deleted jobs, unpublish their products, and get ready to supersede them
+            from product.models import ProductFile
```
wardcr: This block should be in a try/except if you want to maintain the package hierarchy where ProductFile is optional here.

scale/recipe/test/test_models.py
```diff
@@ -543,6 +543,10 @@ def test_successful_supersede_different_recipe_type(self):
         recipe = Recipe.objects.get(id=handler.recipe_id)
         recipe_job_1 = RecipeJob.objects.select_related('job').get(recipe_id=handler.recipe_id, job_name='Job 1')
         recipe_job_2 = RecipeJob.objects.select_related('job').get(recipe_id=handler.recipe_id, job_name='Job 2')
+        job_exe_2 = job_test_utils.create_job_exe(job=recipe_job_2.job)
+        from product.models import ProductFile
+        from product.test import utils as product_test_utils
```
wardcr: This block should be in a try/except if you want to maintain the package hierarchy where ProductFile is optional here.

wardcr:
LGTM

wardcr:
LGTM

PR #357, opened 2016-06-22
Issue 355 - Fix Task Launch Detection

dfaller:
Right now we expect there to not be an exit code when a task launch fails. It appears that Mesos is now returning an exit code of 1 when a task launch fails.

Remove the requirement for no exit code when detecting a task launch. Basically remove

> and not task_results.exit_code:

from line 184 of base_task.py and update applicable unit tests.
Make sure that task launch errors are triggered even when an exit code is returned.

I also had to add checks to make sure the task started running before looking at the exit code to determine the error.


gisjedi:
lgtm

PR #358, opened 2016-06-23
Issue 351 - Job Types Failure Rates updates

### LINKED ISSUE #351: Job Types Failure Rates updates
Ordering for the Job Types Failure Rates page seems to be wrong, in that job types with no failures are interspersed with job types that contain errors.

Also, we need to display the total job count for job types that have no failures instead of just displaying "No Errors".

ericsvendsen-mil:
Resolves #351

Also, this PR addresses job type filtering for the failure rates grid, as well as additional issues related to app-wide data grid layout fixes and filtering persistence.


wardcr:
LGTM as for as addressing the original issue. However, I still feel strongly that we should not try to use a fixed-height table and instead just use the natural vertical scrolling behavior of the browser. The current implementation is more complicated to code and has enough drawbacks that I don't see how it can be preferable to just using standard behavior. If that's agreeable, it can still be a separate issue if you like. For example:

- I still see double scroll bars on most pages.
- What happens if the page size changes?
- What happens if I resize the browser window?
- As the list of job types and records continue to grow, they won't all fit on the screen anyway.

ericsvendsen-mil:
This is an issue with Angular UI grid itself: https://github.com/angular-ui/ui-grid/issues/1735.  I implemented one of the workarounds mentioned in that thread in commit 75388d5, and it seems to be working.  The double scrollbars became more of an issue when the site navigation increased in length, so hopefully this will solve the problem.

wardcr:
Okay this seems to address all of my issues.

However, I'm seeing a very odd bug now. It seems like I'm unable to use the mouse wheel to scroll down while hovering over any part of a table. The wheel works if I hover over any part of the page background and strangely enough, scrolling back up with the mouse wheel does work over the table. Testing with Chrome 51.0.2704.106 m (latest) on Windows 10.

ericsvendsen-mil:
So apparently this is yet another known issue with ui grid.  Working around it is proving to be incredibly annoying, so my recommendation is that we dump the last commit and move the whole grid scrolling problem to a separate issue. There's too much going on here that shouldn't be holding up the other fixes for #351.  Thoughts?

wardcr:
Yeah, that's fine with me. I would prefer to keep the issues small and focused anyway.

PR #361, opened 2016-06-28
Issue 360 - Update S3 Broker to allow role-based access

wardcr:
Changed S3 broker validation to make the credentials attribute optional and updated associated docs and tests.


dfaller:
LGTM, I'll merge.

PR #366, opened 2016-07-01
Issue 362 - Datagrid scrolling

### LINKED ISSUE #362: Datagrid scrolling
Make grids expand to show as many rows as necessary and remove any internal scrolling of the grid.

ericsvendsen-mil:
Fixes #362.

By default, Angular ui-grid forces grids to be a set height with internal scrolling.  This is asinine, since paging prevents the browser from having to load hundreds or thousands of rows at a time.  Here's their guiding philosophy as stated by one of the project's core members:

> the key reasoning behind ui.grid is to deal with large data sets without unreasonable browser overhead. When you generate a very large DOM (think thousands of rows), it puts a lot of load on the browser. With row virtualisation (and therefore, implicitly, with scrolling) the grid only renders those DOM elements that are actually visible. When you scroll the grid we are behind the scenes creating and destroying DOM elements (well, actually we reuse the existing ones) to provide the illusion you're scrolling through a huge list of data that is all rendered and waiting for you.

> Once we go to showing a flat list of everything we'd be bypassing all this logic, and to some extent what we'd then have is just a glorified ng-repeat. Certainly before the project became as full-featured that would have made little sense - it'd be easier to just have an ng-repeat without the overhead that ui-grid adds. I guess now that we have features like exporter, importer, sorting, filtering etc, it may make more sense to use ui.grid in that way, but it certainly isn't the design direction of the project.

So with that in mind, and after extensive research, trial and error, and tinkering, the best solution for now is to remove ui-grid from Bower's control and modify its source to disable the custom scroll handler.  This is not an ideal long-term solution, but too much time has been spent on this already and doing it this way just plain works.

Going forward, we should probably either look into another datagrid solution or spend the time to modify ui-grid so that scroll handling is configurable and/or removed when using server-side paging, and submit a PR to that project.  Based on their philosophy though, I have my doubts that it would ever get merged.  We could also fork their repo, but I'm unsure at this time which Github account would be used to do that.


dickmancj-mil:
LGTM. Manually merged ingestRecords controller and template, chose to remove filename filter from grid header and use the filename filter in section with date filters instead.

PR #368, opened 2016-07-01
Issue 349

dickmancj-mil:
Includes new Ingest Details view with source file details including links to Products and the jobs that produced each product


ericsvendsen-mil:
LGTM

PR #371, opened 2016-07-05
Issue 364 - Upgrade S3 broker to use boto3

wardcr:
This version should now work with any configuration method supported by boto3, such as the EC2 IAM role, AWS credentials file, AWS config file, or environment variables. The existing solution to have hard-coded credentials in the workspace config should also continue to work for ease of testing.

See here for more information:
http://boto3.readthedocs.io/en/latest/guide/configuration.html

I'm not sure if this code will work for custom boto3 builds on other networks yet.


dfaller:
LGTM, I'll go ahead and merge.

PR #375, opened 2016-07-12
Issue 58 - Add Superseded Info to Job Detail API

wardcr:
Implemented according to the original issue, except root_superseded_job_id and superseded_job_id fields are replaced with nested objects per the standard REST conventions.


dfaller:
LGTM

PR #377, opened 2016-07-12
Issue #376 - Datagrids disappear when no data is returned

### LINKED ISSUE #376: Datagrids disappear when no data is returned

ericsvendsen-mil:
Removed the `ng-if="vm.gridOptions.data.length > 0"` statement from all grids.  That was originally there to ensure that they render properly without internal scrollbars, but from my testing it seems as though it's not required to accomplish this.  So when reviewing this PR, just ensure that all the grid rows render as expected, without internal scrollbars, and that page scrolling is not adversely affected in any way.

Both the Jobs and Ingests grids will filter on status when running the static build.

Fixes #376


dickmancj-mil:
Looks good to me.

PR #378, opened 2016-07-13
Issue 59 - Add Superseded Info to Recipe Detail API

wardcr:
Implemented according to the original issue, except root_superseded_recipe_id and superseded_recipe_id fields are replaced with nested objects per the standard REST conventions. Also used the RecipeBaseSerializer with a bit of added info to be consistent with the jobs API.

Made a couple small changes to the recipe handler API to make it more general for additional use cases and included some minor formatting/warning cleanup.


dfaller:
LGTM

Issue #62, opened 2016-03-11
Add Superseded Info to Get Product Updates API


dfaller:
Requires issue #53 to be done first

Filter out any products that have is_superseded=True.

PR #379, opened 2016-07-13
Issue 61 - Add Superseded Info to Get Products API

wardcr:
Added is_published filter parameter and excluded superseded products.

Note this also closes issue #62 because they are using the same model method internally. Although I went ahead and implemented it, I want to confirm this produces the expected behavior for the search scraper use case. If a product is indexed and then later superseded, how does the search index deal with that if the product updates view excludes superseded products?


wardcr:
Confirmed that new versions of products will have the same UUID to allow search indexing to keep in sync, therefore it's safe to exclude the original superseded products from the feed.

dfaller:
LGTM

PR #381, opened 2016-07-14
Issue 313 - Scale Volume Name Uniqueness

wardcr:
Updated scheduler to capture framework ID from Mesos and pass it all the way through the system to the job exe container module so it can be used to construct unique volume names. Fixed some warnings.


dfaller:
LGTM

PR #382, opened 2016-07-15
Fix vagrant

tclarke:
A number of fixes to try and get the standalone vagrant working again.
Includes converting over the rest of the examples to use the embedded job type definitions and some additions to the cli.


dfaller:
LGTM

PR #383, opened 2016-07-18
Issue 380 - added visualization showing products of ingested sources files

dickmancj-mil:
Visualization shows ingested source file on left side of graph, jobs that used it as input and outputs produced from those jobs.

### DIFF
--- a/scale-ui/app/modules/common/services/scaleService.js
+++ b/scale-ui/app/modules/common/services/scaleService.js
@@ -83,6 +83,16 @@
                 }
                 return false;
             },
+            getMediaTypeHtml: function(mediaType){
+                var iconHtml = '<i class="fa ' + 'fa-file' + '"></i>';
+                if(mediaType){
+                    var ic = _.find(scaleConfig.mediaTypes, {mimeType: mediaType})
+                    if(ic){
+                        iconHtml = '<i class="fa ' + ic.icon + '"></i>';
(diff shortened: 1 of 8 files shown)


wardcr:
Visualization looked reasonable to me with the static data.

ericsvendsen-mil:
A couple notes after looking at this with "real" data:

* Some ingest files do not have any associated detail (anything that's queued, but also sometimes even files that have been ingested).  When this happens, the api returns `{"detail":"Not found."}`. This should be accounted for in the UI.
* For files with a lot of jobs and products, the tree visualization is too tall to be really useful, since you need to scroll up and down so often just to see everything.
* It pushes the table below way far down on the page.
* When you do try to scroll to see everything, the page doesn't scroll when the mouse cursor is hovering over the tree view for some reason.
* My initial thought of getting around these issues would be to make the tree collapsible (and be collapsed by default) and possibly putting it in its own modal window.  If it's inline with other DOM elements, the constant resizing due to collapsing/expanding nodes could get tiresome.

Overall it's close, but we need to mock up a more robust static example in order to resolve these problems.

dickmancj-mil:
I think I've addressed all of the issues @ericsvendsen-mil identified. Added more robust test data, resizing graph container to fit the specified viewbox, and allowed scrolling event to pass through to the page.

PR #384, opened 2016-07-19
Issue 161 - Update job views to support filtering by multiple status values

wardcr:
Updated all REST views that have a status filter related to jobs. Only exception was the job re-queue API because it is a potentially breaking change due to it being packaged as a JSON POST rather than query string parameters.


dfaller:
LGTM

PR #385, opened 2016-07-19
Issue 321 - Filter Ingest REST API by Strike

wardcr:
Added strike_id filter parameter to ingests view that accepts multiple values.
Updated status filter to accept multiple values.


dfaller:
LGTM

PR #386, opened 2016-07-19
Issue 283 - Pending Jobs REST API

wardcr:
Added a new endpoint at /job-types/pending/ that basically follows the same conventions as the existing /running/ API.


dfaller:
LGTM

PR #387, opened 2016-07-19
Issue 188 - General algorithm error should not be used on system jobs

wardcr:
Changed job task to provide a default 'unknown' error when a system job fails instead of defaulting to 'algorithm-unknown' error for everything.


dfaller:
I made a quick doc change. Also, we want this to also cover pre and post tasks. You can either pass the 'unknown' default error in the pre and post task files, or just change the default to default_error_name='unknown' (my preference).

dfaller:
Never mind my previous comment. LGTM

PR #388, opened 2016-07-25
Issue 192

gzehring:
Modified the scheduler with the following changes: 

1. Catch any exceptions when the Mesos Driver is run
2. Make sure a shutdown is performed on the thread no matter what status is returned from the Mesos Driver, or if an exception was caught


dfaller:
I removed an extra blank line, but other than that LGTM.

PR #398, opened 2016-07-28
Issue 19 - Strike Refactor

dfaller:
Strike was refactored to a more abstract/general solution that does not require NFS.

There is now a "monitor" base class that abstracts how to watch a workspace for incoming files. The monitor and workspace details are set in the Strike configuration. The Strike configuration allows the previous file name regex matching and allows registering the new files in the monitored workspace or moving the files to a new workspace. The Strike configuration has been updated to schema version 2.0 for these changes.

Two monitors have been implemented so far:
 - our original directory watching monitor
 - an S3 monitor that polls an SQS queue for object creation events in an S3 bucket

This update removes all NFS mounting done directly by Scale and removes the need for any Scale job containers to run as privileged.


scale/docs/architecture/strike.rst
```diff
@@ -4,116 +4,193 @@
...
-to check against the names of newly copied files. If the expression matches a newly copied file name in the directory,
-that file is ingested according to the other fields in the JSON object. The *data_types* field is list of strings. Any
-file that matches the corresponding regular expression will have the data type strings "tagged" with the file. The data
-type tags are used to categorize files and control which jobs and recipes they go to. The *workspace_path* specifies a
-relative path within the workspace where each file will be stored. Three additional and dynamically named directories,
-for the current year, month, and day, will be appended to the *workspace_path* value automatically by the Scale system
-when a file is ingested (i.e. workspace_path/YYYY/MM/DD). The *workspace_name* field is the system name of the unique
-workspace that should ingest the file. To see all of the options for a Strike process's configuration, please refer to
-the Strike Configuration Specification below.
+When a new file is copied into the monitored workspace, its file name is checked against a number rules using regular
```
tclarke: Typo? Should probably be "a number of rules"

scale/docs/architecture/strike.rst
```diff
@@ -4,116 +4,193 @@
...
+    version, which is currently 2.0. It is recommended, though not required, that you include the *version* so that
     future changes to the specification will still accept your Strike configuration.
 
-**mount**: JSON string
+**workspace**: JSON string
 
-    The *mount* field is a required string that specifies the NFS host and path that should be mounted in order to
-    access the monitored directory (format is *host:/file/path*).
+    The *workspace* field is a required string that specifies the name of the workspace that is being monitored. The
+    type of the workspace (its broker type) will determine how incoming files can be monitored.
```
tclarke: Is this true? Below it suggests the **monitor** field determines how things are monitored. This should be clarified.

scale/ingest/test/test_views.py
```diff
@@ -589,5 +593,6 @@ def test_warnings(self):
         results = json.loads(response.content)
 
         self.assertEqual(response.status_code, status.HTTP_200_OK, response.content)
-        self.assertEqual(len(results['warnings']), 1)
-        self.assertEqual(results['warnings'][0]['id'], 'mount_change')
+        # TODO: update for new validation
```
tclarke: Does this need to be addressed now?

tclarke:
The CLI needs to be updated to handle the new strike schema. Look at strike.go in the top level of the CLI.

tclarke:
Once above issues are addressed, LGTM

scale/ingest/test/test_views.py
dfaller: I removed the unit test. With our new monitors we don't currently have any warnings.

tclarke:
LGTM

PR #400, opened 2016-08-01
Issue 218 - Job Configuration Documentation

dfaller:
I added documentation on the job configuration schema that was added as part of the Scale Dockerization.

I did not remove the job_environment field on the job_execution model since this would impact the REST API and UI. It can be removed in the future with other job_execution changes.


tclarke:
LGTM

PR #403, opened 2016-08-01
Issue 391 - Mesos.interface could not be imported for new development environments

gzehring:
mesos.interface library is now being pulled from pip

mesos.interface imports are no longer inside of a try catch block, and no longer falling back on stubs


gzehring:
Unintentionally merged, but now reverted and waiting for code check before merge again.

PR #405, opened 2016-08-01
Issue 391 - Mesos.interface errored on import for new development environments

gzehring:
Added mesos.interface to the list of requirements in dev_win.txt

Moved the mesos.interface imports out of the try/catch block in all occurrences, no longer falling back on stubs

### DIFF
--- a/scale/scheduler/management/commands/scale_scheduler.py
+++ b/scale/scheduler/management/commands/scale_scheduler.py
@@ -8,25 +8,14 @@ from optparse import make_option
 
 from django.conf import settings
 from django.core.management.base import BaseCommand
+from mesos.interface import mesos_pb2
+from pesos.scheduler import PesosSchedulerDriver as MesosSchedulerDriver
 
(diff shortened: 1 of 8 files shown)


scale/scheduler/management/commands/scale_scheduler.py
```diff
@@ -8,25 +8,14 @@
...
 
-# Try to import production Mesos bindings, fall back to stubs
-try:
-    from mesos.interface import mesos_pb2
-    from pesos.scheduler import PesosSchedulerDriver as MesosSchedulerDriver
-    logger.info('Successfully imported pesos bindings')
-except ImportError:
-    try:
-       from mesos.interface import mesos_pb2
-       from mesos.native import MesosSchedulerDriver
```
dfaller: You need to keep the import at line 24 instead of the import at line 19.

gzehring: Done!

dfaller:
LGTM, you can merge and delete the branch.

PR #415, opened 2016-08-04
Issue 15 - new logging interface

### LINKED ISSUE #15: Refactor log system
Refactor the log system:
Remove all of the Django database logging (LogEntry model in Error app)
No longer include task logs in the database
Figure out a new method of log storage so that logs don't need to be moved once a task finishes
Store task logs independently (pre/job/post tasks)
See if stdout and stderr can be both combined and separated
Remove stdout/stderr columns in the database and figure out new columns for handling URLs that can be used for log viewing.
Update the JobExecutionManager.task_started() and JobExecutionManager.task_ended() methods and update the tasks to remove stdout/stderr and the URLs
Should this use the workspace system?

This is an important issue for improving the reliability and performance of the scheduler since logs will no longer have to be dealt with during status updates.

tclarke:
closes #15 
Added ELK stack for logging and associated REST API.

The new stack will startup logstash automatically on DC/OS installs but the user must manually create an Elasticsearch cluster. This is because ES has a lot of variables in a cluster setup and it would be unwieldy to automate this. ES has a DC/OS config so installation is easy.

Docker stdout/stderr are logged to logstash (via gelf2) which forwards to Elasticsearch for storage tagging log lines with the name of the job (42_job, etc.) for easy retrieval.

New REST endpoints are available for stdout/stderr/combined output and can be requested in plain text, html, or raw json. The existing API has been kept for accessing data from old job executions but should be considered deprecated.

I also added the DC/OS configs for scale so we can version them along with scale. They will not work from this repo as the DC/OS universe repo layout is required. Before using them, you'll need to clone and setup that repo per the DC/OS universe development instructions.

Removed/deprecated the old logging interface and code.


dickmancj-mil:
Is Elastic Search required or is it still an option to log to postgres?

tclarke:
It's required

scale/docs/rest/job_execution.rst
```diff
@@ -264,10 +264,6 @@ These services provide access to information about "all", "currently running" an
 +----------------------+-------------------+--------------------------------------------------------------------------------+
 | results              | JSON Object       | An interface description for all the possible job results meta-data.           |
 +----------------------+-------------------+--------------------------------------------------------------------------------+
-| current_stdout_url   | URL               | The URL of the standard output log for the execution.                          |
-+----------------------+-------------------+--------------------------------------------------------------------------------+
-| current_stderr_url   | URL               | The URL of the standard error log for the job execution.                       |
-+----------------------+-------------------+--------------------------------------------------------------------------------+
```
dfaller: We should keep these fields in the documentation until they are actually removed from the API. You could just add a "(deprecated)" to the field descriptions.

tclarke: Those have been removed, the stout/stderr fields are still there

scale/job/management/commands/scale_post_steps.py
```diff
@@ -106,6 +106,7 @@ def _perform_post_steps(self, job_exe):
 
         job_interface = job_exe.get_job_interface()
         job_data = job_exe.job.get_job_data()
+        # TODO: fix this
```
dfaller: We need to pull the new logs (for the job task) in order to do stdout product registration. We still have jobs running that rely upon this.

scale/job/models.py
```diff
@@ -1085,10 +1062,6 @@ def task_ended(self, job_exe_id, task, when, exit_code, stdout, stderr):
         :type when: :class:`datetime.datetime`
         :param exit_code: The task exit code, possibly None
         :type exit_code: int
-        :param stdout: The task stdout contents, possibly None
-        :type stdout: string
-        :param stderr: The task stderr contents, possibly None
-        :type stderr: string
         """
 
         # TODO: once we no longer have stdout and stderr in job execution model, transition this from selecting with
```
dfaller: Resolve this TODO by making task_ended() use an update query just like task_started() does. The locking was only needed when appending stdout/stderr to existing stdout/stderr contents which we are no longer doing. The update is faster and cleaner.

scale/job/management/commands/scale_post_steps.py
tclarke: Oops, forgot to add that.. I'll get that done soon

scale/job/models.py
```diff
@@ -1499,35 +1461,74 @@ def uses_docker(self):
...
-        :rtype: string
-        """
-        if stderr:
-            if self.stderr:
-                self.stderr += stderr
-            else:
-                self.stderr = stderr
-        return self.stderr
+    def get_log_json(self, include_stdout=True, include_stderr=True, since=None):
+        '''Get log data from elasticsearch as a dict (from the raw JSON).
```
dfaller: Should use triple double-quotes for docstrings (""")

dfaller:
All of my comments are currently in. Just a couple things, the mains ones being grabbing logs in the post task for the old stdout method of registering products and fixing task_ended() so it doesn't do expensive query lock anymore.

dfaller:
Your updates to my comments look good. Let's have you work with one of the of the UI guys next week to get the bare minimum of the UI working with the new log APIs. We don't want to merge and have log viewing being broken.

PR #417, opened 2016-08-05
Issue 370 - Passing Docker Credentials to Mesos

gzehring:


scale/scale/local_settings_dev.py
```diff
@@ -56,3 +56,6 @@
 
 # The full name for the Scale Docker image (without version tag)
 SCALE_DOCKER_IMAGE = 'geoint/scale'
+
+# The location of the config file containing Docker credentials
+CONFIG_URI = os.environ.get('CONFIG_URI', CONFIG_URI)
```
tclarke: We don't generally use env vars in the dev settings. Since there's no reasonable default for this, it should default to settings.py so just remove this change.

scale/scale/settings.py
```diff
@@ -29,6 +29,9 @@
 # The full name for the Scale Docker image (without version tag)
 SCALE_DOCKER_IMAGE = 'geoint/scale'
 
+# The location of the config file containing Docker credentials
+CONFIG_URI = os.environ.get('CONFIG_URI', None)
```
tclarke: Don't use env vars in the settings.py. This should be a default value only so change this to `CONFIG_URL = None`

scale/mesos_api/tasks.py
```diff
@@ -103,6 +104,7 @@ def _create_docker_task(task):
     arguments = task.command_arguments.split(" ")
     for argument in arguments:
         mesos_task.command.arguments.append(argument)
+    mesos_task.command.arguments.append(settings.CONFIG_URI)
```
tclarke: It's unclear if this will work with an http, etc. URI since it only talks about file URIs in the documentation. Files aren't an option since the users aren't generally going to be able to place the config on each node (especially in aws type environments where nodes may come and go at any time).

scale/scale/settings.py
tclarke: You also need to change the dockerfile and entrypoint script so this can be passed when executing the docker container. In addition, you'll probably need to update the DC/OS deployment configuration to include a DC/OS option to set this value. issue-15 (when it's merged) checks in a copy of the DC/OS config where you can make the change.

scale/mesos_api/tasks.py
gzehring: After talking with Derick earlier I don't believe this is where the CONFIG_URI needs to be assigned. Still working on this. I've found a couple of different examples, but they contradict each other. Http should work with this as well based on what I am seeing. 

One Marathon example of this: 
[mesos_task.command.arguments.append(settings.CONFIG_URI)](mesos_task.command.arguments.append(settings.CONFIG_URI))

scale/scale/settings.py
```diff
@@ -29,6 +29,10 @@
 # The full name for the Scale Docker image (without version tag)
 SCALE_DOCKER_IMAGE = 'geoint/scale'
 
+# The location of the config file containing Docker credentials
+# The URI value should be something like this 'http://marathon-ip:8080/v2/artifacts/.dockercfg'
```
tclarke: Marathon artifact store isn't available on a default DC/OS or Marathon deployment and is currently pretty limited in use. Suggest the following change:
`# The URI value should point to an externally hosted location such as a webserver or hosted S3 bucket. The value will be an http URL such as 'http://static.mysite.com/foo/.dockercfh`

tclarke:
LGTM

PR #423, opened 2016-08-10
Issue 287 - Add Strike management UI

### LINKED ISSUE #287: Add Strike management UI
Now that we have new REST APIs for listing, creating, and updating Strike processes it would be good to make that available in the UI.

ericsvendsen-mil:
Closes #287


ericsvendsen-mil:
Changes that need to be made:

* "Monitor Type" field depends on the type of workspace chosen. If an S3 workspace is selected, then an S3 monitor type will be required.  If a Host workspace is selected, then a dir-watcher monitor type will be required.  If an NFS workspace is selected, then no valid monitor type will be available.
* Change "Strike Ingest Files" to "Ingest File Rules", "Current Files" to "Current Rules", and "Add File to Configuration" to "Add Rule to Configuration"
* Strike version should not be editable or displayed - always send "2.0" when saving a strike
* "New Workspace" should be a dropdown.  Exclude the workspace that is selected as the workspace for the strike.
* Only "Filename Regex" is required when adding a rule
* Move the strikes UI to a section under Data Feed

dickmancj-mil:
Moved Strikes UI under Data Feed. This completes the work identified on Strikes UI, ready for final review before merging.

ericsvendsen-mil:
Found a bug where the workspace dropdown was assuming it would be dealing with a workspace object, when in reality only the workspace name is returned by the api.  Should be good to go now, pending further review by @dickmancj-mil (or whoever gets to this first).

PR #424, opened 2016-08-12
Issue 418 - Fix Node Detail API

gzehring:
1. - Removed PUT method
2. - Removed unit tests for PUT method
2. - Modified PATCH method
      - Can no longer update hostname or port
      - Now returns node details and resources
3. Modified unit tests for PATCH method to verify we are returning node details and resources


scale/node/views.py
```diff
@@ -142,9 +110,30 @@ def patch(self, request, node_id):
...
+            slave_info = mesos_api.get_slave(sched.master_hostname, sched.master_port, node.slave_id, True)
+            if slave_info and slave_info.total:
+                resources = slave_info.to_dict()['resources']
+        except:
+            logger.exception('Unable to fetch slave resource usage')
+        if resources:
+            result['resources'] = resources
+        else:
+            result['disconnected'] = True
+
```
dfaller: Since all of this stuff is identical to retrieving/processing the node detail that is in the GET method, factor both of these out into a helper method called something like _get_node_details() for the view. That way the logic to return node details is not duplicated.

gzehring: Yup sounds like a good idea to me

dfaller:
I think the last thing is to update the REST API docs with your changes in scale/docs/rest/node.rst.

gzehring:
Documentation has now been updated to reflect the changes

scale/docs/rest/node.rst
```diff
@@ -288,41 +288,40 @@ These services provide access to information about the nodes.
...
+|                    |                   | for execution. Remaining tasks for a previously executing job will complete.   |
 +--------------------+-------------------+--------------------------------------------------------------------------------+
 | **Successful Response**                                                                                                 |
 +--------------------+----------------------------------------------------------------------------------------------------+
-| **Status**         | 201 CREATED                                                                                        |
+| **Status**         | 200 OK                                                                                        |
 +--------------------+----------------------------------------------------------------------------------------------------+
-| **Location**       | URL pointing to the details for the node (should be the same as the request URL)                   |
+| **Location**       | URL pointing to the details for the node (should be the same as the request URL).                  |
```
dfaller: This should be removed. A Location header should only be returned with a 201 response.

scale/docs/rest/node.rst
```diff
@@ -337,75 +336,142 @@ These services provide access to information about the nodes.
...
+|           "used": {                                                                                                     |
+|               "cpus": 16.0,                                                                                             |
+|               "mem": 63305.0,                                                                                           |
+|               "disk": 131485.0                                                                                          |
+|           }                                                                                                             |
+|       }                                                                                                                 |
+|   }                                                                                                                     |
 +--------------------+-------------------+--------------------------------------------------------------------------------+
 | **Error Responses**                                                                                                     |
```
dfaller: Remove this "Error Responses" section. We don't use this anywhere else in the REST docs.

scale/node/views.py
```diff
@@ -98,55 +75,50 @@ def put(self, request, node_id):
...
-        :type node_id: str
-        :rtype: :class:`rest_framework.response.Response`
-        :returns: the HTTP response to send back to the user
-        """
-
-        extra = filter(lambda x, y=self.update_fields: x not in y, request.data.keys())
-        if len(extra) > 0:
-            return Response('Unexpected fields: %s' % ', '.join(extra), status=status.HTTP_400_BAD_REQUEST)
+        return Response(result, status=status.HTTP_200_OK,
+                        headers={'Location': request.build_absolute_uri()})
```
dfaller: Remove the "Location" header. Location headers are to be used for 201 Responses. Returning Response(result) might default to using the 200 OK status code.

gzehring:
The recommended changes are completed!

dfaller:
LGTM

PR #431, opened 2016-08-18
Issue 129 - Add two new exit codes to post steps

gzehring:
Added the following exit codes and unit tests for them:

IV_MF_CODE - Invalid Manifest
MI_OP_CODE - Missing Manifest Output


scale/error/models.py
```diff
@@ -59,6 +59,22 @@ def get_nfs_error(self):
...
+        return self.get_builtin_error('invalid-results-manifest')
+
+    def get_missing_output_error(self):
+        """Returns the error for missing a required output
+
+                :returns: The missing output error
+                :rtype: :class:`error.models.Error`
+                """
+        return self.get_builtin_error('missing-required-output')
+
```
dfaller: The new errors should be in the job app fixture file for basic job errors. The fixture files are how these builtin errors get populated in the database. The get_invalid_manifest_error() and get_missing_output_error() methods can be removed. The new job errors can simply be accessed in the job app using get_builtin_error().

scale/job/management/commands/scale_post_steps.py
```diff
@@ -71,6 +76,10 @@ def handle(self, **options):
                 exit_code = NFS_EXIT_CODE
             elif isinstance(ex, IOError):
                 exit_code = IO_EXIT_CODE
+            elif isinstance(ex, InvalidResultsManifest):
```
dfaller: The invalid results manifest exception should also be thrown when file paths within the results manifest are invalid, not just when the JSON schema itself is invalid.

scale/job/management/commands/scale_post_steps.py
```diff
@@ -71,6 +76,10 @@ def handle(self, **options):
...
+                exit_code = IV_MF_CODE
+            elif isinstance(ex, ResultsManifestAndInterfaceDontMatch):
```
dfaller: We want to get rid of the ResultsManifestAndInterfaceDontMatch exception as it is confusing and doesn't tell the user what the problem is. We need a MissingRequiredOutput exception instead. Also improve the logging so that if a required output is not provided by the manifest, the details (which output was required, etc) are logged. Also create any additional exceptions/logging needed to cover all error cases currently handled by ResultsManifestAndInterfaceDontMatch (such as output type mismatches).

PR #435, opened 2016-08-18
Can now edit 'is_paused' field for a system job

gzehring:
Modified the IF statement that was throwing an exception if a system job was attempted to be modified. Now instead of always throwing an exception for an attempted edit on a system job, an exception is only thrown if a field other than 'is_paused' is attempted to be edited.

Also added a unit test to verify this

### DIFF
--- a/scale/job/test/test_models.py
+++ b/scale/job/test/test_models.py
@@ -996,21 +996,19 @@ class TestJobTypeManagerEditJobType(TransactionTestCase):
 
         name = 'my-job-type'
         version = '1.0'
-        is_paused = False
-        new_is_paused = True
         trigger_rule = trigger_test_utils.create_trigger_rule(trigger_type=job_test_utils.MOCK_TYPE,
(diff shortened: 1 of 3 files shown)


scale/job/test/test_models.py
```diff
@@ -991,6 +991,25 @@ def test_system_job_type(self):
...
+        job_type.is_system = True
+        job_type.save()
+
+        # Call test
+        JobType.objects.edit_job_type(job_type.id, is_paused=True)
+
+        # Check results
+        job_type = JobType.objects.select_related('trigger_rule').get(pk=job_type.id)
+        # No change
+        self.assertEqual(job_type.is_paused, True)
```
dfaller: If this works correctly, then is_paused should be false here.

dfaller:
Make sure that the REST API returns a 400 Bad Request if the user tries to edit a different field for a system job type. You can use the InvalidJobField exception for this.

dfaller:
LGTM

Issue #411, opened 2016-08-03
UI API prefix version


ericsvendsen-mil:
Hardcode the version to each API call in the UI (e.g. `v3`).  That way when new API versions are introduced, the UI can be upgraded incrementally.

dickmancj-mil:
For a balance of configurability, maintenance and performance, considering supporint ability to overrides specific top-level REST service endpoints. Example:
```
"urls": {
            "apiPrefix": "/api/",
            "overrides": {
                "ingests": "/api/v3/"
            },
            "documentation": "/docs"
        },
```
This overrides apiPrefix for all calls to the "<api-root>/ingests" endpoints including /ingests/ and /ingests/<id>/  This wouldn't support overriding only one or the other.  Would this be enough configurability @dfaller ?

PR #436, opened 2016-08-19
Issue #411 - UI API prefix version

dickmancj-mil:
Pull request for Issue #411. Implemented ability to override API prefix config for 1st level REST endpoints. Included sample config in scaleConfig.json that overrides prefix for /ingests/* calls.


tclarke:
LGTM

PR #439, opened 2016-08-19
Fixes #428, requeue and cancel job buttons have been fixed

### LINKED ISSUE #428: Fix Recipe trigger editor
Recipe Trigger data_types field does not work because the UI template is binding to the wrong field.

dickmancj-mil:


tclarke:
LGTM

PR #445, opened 2016-08-23
Overview margin fix

ericsvendsen-mil:
Removed a legacy css class that was causing a gap to appear on the overview page.


dickmancj-mil:
LGTM

PR #450, opened 2016-08-26
Fix missing exit code

### LINKED ISSUE #359: Scale Docker Image Exit Code
We don't appear to be capturing the exit codes for system jobs and pre/post tasks. I think the issue is because we are not capturing and returning the exit code from the manage.py call in the entrypoint script.

tclarke:
closes #359 
If there's no exit code on a task finished, set it to 0.

### DIFF
--- a/scale/scheduler/scale_scheduler.py
+++ b/scale/scheduler/scale_scheduler.py
@@ -255,9 +255,9 @@ class ScaleScheduler(MesosScheduler):
                 if status.state == mesos_pb2.TASK_RUNNING:
                     running_job_exe.task_running(task_id, results.when)
                 elif status.state == mesos_pb2.TASK_FINISHED:
-                    running_job_exe.task_complete(results)
-                    if results.exit_code is None or not type(results.exit_code) is type(int):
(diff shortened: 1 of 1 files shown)


scale/scheduler/scale_scheduler.py
```diff
@@ -256,6 +256,8 @@ def statusUpdate(self, driver, status):
                     running_job_exe.task_running(task_id, results.when)
                 elif status.state == mesos_pb2.TASK_FINISHED:
                     running_job_exe.task_complete(results)
+                    if results.exit_code is None or not type(results.exit_code) is type(int):
+                        results.exit_code = 0
```
dfaller: This should be before `running_job_exe.task_complete(results)`

I don't think the int type check is really needed. Shouldn't it be "is not"?

tclarke: I wasn't sure about the type guarantee for the exit_code. I dug into the mesos code a bit more and it looks like it's going to be an int if it's there so I think we're safe removing the check.

dfaller:
LGTM

Issue #430, opened 2016-08-17
Add fields to Job Details


dickmancj-mil:
Requested fields are:

1. Duration (same as on Jobs list)
2. If executions, Node of last execution
3. If executions, Exit code of last execution

dfaller:
I would recommend that this replace the current timeline section of the job detail page. It can have a breakdown of each task from the most recent job execution with the start, stop, and exit code of each task. Should also include queued time/duration for the most recent job execution.

dfaller:
PR #451 closes #430

PR #451, opened 2016-08-26
Issue #430 - Add fields to Job Details

dickmancj-mil:
Added Duration to Job Details. Added Job Exit Code and a link to Node on the Executions table.


tclarke:
LGTM

cshamis:
On the job details page, can you add a button or linkback to the "product hierarchy" ?  

Also maybe even a button or link to the "recipe instance" ... is that doable?

--Then it becomes possible to "walk" a recipe up and down...

PR #452, opened 2016-08-29
Issue #60 - wired up /api/ingests file_name filter to Search box

### LINKED ISSUE #60: Walking ingest record to recipe instance
A frequent use-case that occurs is being able to follow ingested files through to the recipes that ran, and (eventually) down to the execution logs.  This is a fairly common task, and is not accomplished easily at present.  

It would be very helpful to have a filename searchbar that filtered the ingest record(s) and also if ingest-records could be made "clickable" to take you directly to the recipe.  Since there may be more than one recipe, a popup with a list of "Which recipe do you want?"  might be in order...

dickmancj-mil:
Closes #60 

Wired Search box on Data Feed/Ingests to filter Ingest records on file_name


ericsvendsen-mil:
Should the file_name filter value be stored in the url querystring like other filter values? Otherwise LGTM.

ericsvendsen-mil:
Added filename to url...

Issue #434, opened 2016-08-18
change If-Modified-Since behavior in log API


tclarke:
The If-Modified-Since header is also being used for partial log retrieval. It should only check to see if an updated log should be sent or not. Use a query param to specify the 'since' time for partial retrieval. This is causing confusion when directly accessing the log API from a browser which will often insert the If-Modified-Since automatically.

PR #454, opened 2016-08-29
#434 Changed If-Modified-Since behavior

tclarke:
If-Modified-Since returns the requested data if there's new data since. "started" param now controls which data to return.


scale/job/views.py
```diff
@@ -699,18 +699,25 @@ def retrieve(self, request, job_exe_id, log_id):
...
             since = util.parse.datetime.datetime.strptime(since, "%a, %d %b %Y %X %Z")
         else:
             since = None
 
+        if since is not None and started != since:
+            logs, last_modified = job_exec.get_log_json(include_stdout, include_stderr, since)
+            if logs is None:
+                rsp = HttpResponse(status=304)
+                rsp["Last-Modified"] = last_modified.strftime("%a, %d %b %Y %X %Z")
+                return rsp
```
dfaller: Having the two different timestamps (since, started) seems overly complicated and requires an extra ES API call. I would recommend that we just don't use IF_MODIFIED_SINCE.

tclarke: If-Modified-Since is understood by HTTP proxies, caches, etc. so it's necessary for proper function of HTTP intermediates

dfaller: It's not really necessary as intermediary caching is not important for this. The header over complicates the REST logic while requiring a second ES API call. Using only the query parameter is simpler and faster.

gisjedi:
LGTM. 🚢

PR #455, opened 2016-09-01
Issue 412 - Initial Scheduler Refactor

dfaller:
This issue was a short-term fix to improve scheduler performance and database all status updates from Mesos. I did the following:

1. Moved database queries pertaining to each task update into a single query performed when a job execution was finished (completed/failed).
2. Added a new scheduler background thread to database all Mesos task status updates that we receive.
3. Various bug fixes for logging
4. Key bug fix for reporting errors from pre and post tasks


scale/job/execution/running/tasks/base_task.py
```diff
@@ -18,15 +18,20 @@ def __init__(self, task_id, job_exe):
...
+        :type task_id: string
         :param job_exe: The job execution, which must be in RUNNING status and have its related node, job, job_type, and
             job_type_rev models populated
         :type job_exe: :class:`job.models.JobExecution`
         """
 
         self._task_id = task_id
-        self._task_name = 'Job Execution %i (%s)' % (job_exe.id, job_exe.get_job_type_name())
-        self._is_running = False
+        self._task_name = '%s %s' % (job_exe.job.job_type.title, job_exe.job.job_type.version)
```
tclarke: Is this going to cause any issues elsewhere? Are we expecting "Job Execution..." as the name explicitly anywhere? I don't think we are but I'm worried we might be. :-/

scale/mesos_api/utils.py
```diff
@@ -3,11 +3,107 @@
 from datetime import datetime, timedelta
 
 from django.utils.timezone import utc
+from google.protobuf.internal import enum_type_wrapper
```
tclarke: *twitch*

scale/queue/models.py
```diff
@@ -393,7 +399,11 @@ def handle_job_failure(self, job_exe_id, when, error=None):
             # If this job execution is no longer running, ignore failure
             return
         job_exe.job = Job.objects.get_locked_job(job_exe.job_id)
+        for task in tasks:
+            task.populate_job_exe_model(job_exe)
         JobExecution.objects.update_status([job_exe], 'FAILED', when, error)
+        # TODO: extra save here to capture task info, re-work this as part of the architecture refactor
```
tclarke: This note should be added to the appropriate github issue.

tclarke:
LGTM (assuming comments are addressed without any changes needed)

scale/job/execution/running/tasks/base_task.py
dfaller: No, there are no (and shouldn't be) any dependencies on the task name. It's for display purposes only. The new task name is more helpful and descriptive in a DC/OS environment with other tasks/frameworks running.

scale/mesos_api/utils.py
dfaller: Yeah, but it was the cleanest way available to me. We might need to rework this if mesos.interface changes. It would be nice if their Python bindings presented a consistent enum/class/something for us to interface with.

scale/queue/models.py
dfaller: Done.

PR #458, opened 2016-09-02
Issue 456 - Bug with CONFIG_URI set to None

dfaller:
This is a bug fix that checks that the settings.CONFIG_URI is set from the environment variable before setting it to the Mesos task uris field.


tclarke:
LGTM

PR #461, opened 2016-09-03
Issue 460 - Create Strike UI spins endlessly... errors on console

### LINKED ISSUE #460: Create Strike UI spins endlessly... errors on console
When attempting to use the new Strikes configuration in the Scale UI, it hangs on validate with a waiting indicator.  Looking at the console a number of errors are evident and it appears the Javascript is broken prior to the API XHR.

Any minimal allowed form entry causes this issue as of https://github.com/ngageoint/scale/commit/7515d64c83636f6b3f66c53644725312ca320d70

ericsvendsen-mil:
Fixes #460


ericsvendsen-mil:
Added another fix to remove empty values from files_to_ingest objects.

Issue #442, opened 2016-08-22
Improve Logging Framework Robustness


dfaller:
Right now in the logging deployment, Scale is tied to specific node instances for logstash and ElasticSearch. If these nodes go down, Scale's logging capability breaks.

1. Update Scale so that the main logstash and ElasticSearch URLs are hit on a regular basis to retrieve node URLs so that Scale is using up-to-date node URLs. Take care so that the additional API calls don't incur major performance problems.

2. Fix the ability to send multiple ElasticSearch URLs to the scale-logstash config. If you send a list of URLs, the config is invalid (it sees "http://127.0.0.1, http://127.0.0.2" as one invalid URL) and logstash exits with an error.

3. Marathon 1.8 runs on https. We need to try to hit the Marathon API over https and fall back to http if it fails.

PR #470, opened 2016-09-15
#442 - Logstash support for Elasticsearch node failover.

gisjedi:
- Created watchdog process that can manipulate logstash config with new / lost Elasticsearch tasks.
- Created dockerfile that has supervisor managing both Logstash and watchdog process.
- Replaced logstash inline config with a template file that is sourced for injecting Elasticsearch hosts.
- Added environment variable values for remote templates, custom watchdog sleep timing and Elasticsearch framework address.


gisjedi:
Addresses some of bullet 1 and fully bullet 2 of #442.

dockerfiles/framework/scale/deployDb.py
```diff
@@ -22,6 +21,10 @@ def run():
     else:
         cfg['volumes'] = '{"containerPath": "/var/lib/pgsql/data","hostPath": "%(dbHostVol)s","mode": "RW"}' % cfg
 
+    if os.environ.get('INIT_DB', False):
+        # attempt to delete an old instance..if it doesn't exists it will error but we don't care so we ignore it
+        requests.delete('http://marathon.mesos:8080/v2/apps/' + cfg['scaleDBHost'])
+
```
dfaller: Why do we have this code block?

gisjedi: I didn't see value in doing this on INIT_DB or DEPLOY_DB being `True`. Feel free to argue otherwise, but I don't really anticipate a normal user setting DEPLOY_DB to True and remembering to set it to False between restarts or scaling in Marathon. If so, your database would get wiped. 😦  As a result I removed this entire block. The DB connections are already being maintained between scheduler restarts, so the issue that affected log stash doesn't apply here.

PR #471, opened 2016-09-16
Issue 457 - Add API to reprocess a recipe

wardcr:
- Added a new view that supports reprocessing an existing recipe.
- Added job_names and all_jobs filter parameters per original issue.
- Added model method to support reprocessing a single recipe.
- Added tests and docs.
- Changed some generic exceptions to more specific types.

### DIFF
--- a/scale/docs/rest/recipe.rst
+++ b/scale/docs/rest/recipe.rst
@@ -417,10 +417,26 @@ These services provide access to information about recipes.
 | **Reprocess Recipe**                                                                                                    |
 +=========================================================================================================================+
 | Creates a new recipe using its latest type revision by superseding an existing recipe and associated jobs.              |
+| Note that if the recipe type definition has not changed since the recipe was created, then one or more job names must be|
+| specified to force the recipe to be reprocessed.                                                                        |
 +-------------------------------------------------------------------------------------------------------------------------+
 | **POST** /recipes/{id}/reprocess/                                                                                       |
 |          Where {id} is the unique identifier of an existing model.                                                      |
-+-------------------------------------------------------------------------------------------------------------------------+
++--------------------+----------------------------------------------------------------------------------------------------+
+| **Content Type**   | *application/json*                                                                                 |
++--------------------+----------------------------------------------------------------------------------------------------+
+| **JSON Fields**                                                                                                         |
++--------------------+-------------------+----------+---------------------------------------------------------------------+
+| job_name           | String            | Optional | The name of a job within the recipe definition that should be       |
+|                    |                   |          | included in the reprocessing request, even when the definition for  |
+|                    |                   |          | the job has not changed between recipe type revisions.              |
+|                    |                   |          | Duplicate it to filter by multiple values.                          |
++--------------------+-------------------+----------+---------------------------------------------------------------------+
+| all_jobs           | Boolean           | Optional | A flag that indicates all jobs in the recipe should be reprocessed, |
+|                    |                   |          | even when the recipe type definitions are identical. This option    |
+|                    |                   |          | overrides any job_name parameters and is typically used to re-run   |
+|                    |                   |          | jobs that failed due to temporary errors.                           |
++--------------------+-------------------+----------+---------------------------------------------------------------------+
 | **Successful Response**                                                                                                 |
 +--------------------+----------------------------------------------------------------------------------------------------+
 | **Status**         | 201 CREATED                                                                                        |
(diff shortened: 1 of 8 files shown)


scale/docs/rest/recipe.rst
```diff
@@ -409,3 +409,259 @@ These services provide access to information about recipes.
...
+| job_names          | Array[String]     | Optional | The name of jobs within the recipe definition that should be        |
+|                    |                   |          | included in the reprocessing request, even when the definition for  |
+|                    |                   |          | those jobs has not changed between recipe type revisions.           |
++--------------------+-------------------+----------+---------------------------------------------------------------------+
+| all_jobs           | Boolean           | Optional | A flag that indicates all jobs in the recipe should be reprocessed, |
+|                    |                   |          | even when the recipe type definitions are identical. This option    |
+|                    |                   |          | overrides any job_name parameters and is typically used to re-run   |
+|                    |                   |          | jobs that failed due to temporary errors.                           |
++--------------------+-------------------+----------+---------------------------------------------------------------------+
```
dfaller: I wouldn't say "is typically used to re-run jobs that failed due to temporary errors", as that is the point of re-queuing.

scale/docs/rest/recipe.rst
```diff
@@ -409,3 +409,259 @@ These services provide access to information about recipes.
...
+|    {                                                                                                                    |
+|        "all_jobs": true                                                                                                 |
+|    }                                                                                                                    |
++-------------------------------------------------------------------------------------------------------------------------+
+| **Successful Response**                                                                                                 |
++--------------------+----------------------------------------------------------------------------------------------------+
+| **Status**         | 201 CREATED                                                                                        |
++--------------------+----------------------------------------------------------------------------------------------------+
+| **Content Type**   | *application/json*                                                                                 |
```
dfaller: I think in other API docs, we listed the Location header and explained what the URL pointed to.

scale/recipe/views.py
```diff
@@ -308,3 +309,39 @@ def retrieve(self, request, recipe_id):
...
+            raise Http404
+        except ReprocessError:
+            raise BadParameter('Job names must be requested when the recipe type has not changed')
+
+        try:
+            new_recipe = Recipe.objects.get_details(handler.recipe.id)
+        except Recipe.DoesNotExist:
+            raise Http404
+
+        url = urlresolvers.reverse('recipe_type_details_view', args=[new_recipe.id])
```
dfaller: Shouldn't this be recipe_details_view?

scale/recipe/models.py
```diff
@@ -324,6 +333,64 @@ def get_details(self, recipe_id):
...
+
+        # Acquire model locks
+        superseded_recipe = Recipe.objects.select_for_update().get(pk=recipe_id)
+        prev_jobs = Job.objects.select_for_update().filter(pk__in=[rj.job_id for rj in prev_recipe_jobs])
+        prev_jobs_dict = {j.id: j for j in prev_jobs}
+        superseded_jobs = {rj.job_name: prev_jobs_dict[rj.job_id] for rj in prev_recipe_jobs}
+
+        # Create an event to represent this request
+        description = {'recipe_id': recipe_id, 'job_names': job_names}
+        event = TriggerEvent.objects.create_trigger_event('REPROCESS_RECIPE', None, description, timezone.now())
```
dfaller: I think we should use the same "user" trigger event as we do for the REST APIs for queuing new jobs and recipes. The trigger event is not so much about what we are doing (re-processing), but about what caused this to happen (user/API call, a Scale Strike process, etc.) All user-initiated events (API calls) should be grouped together.

scale/docs/rest/recipe.rst
wardcr: Good point, reworded to focus on the algorithm changes use case.

scale/docs/rest/recipe.rst
wardcr: True, although it turns out it was pretty inconsistent. I copied from one that didn't have it which is why I missed it. I went back through and added it to several places that do a 201.

scale/recipe/views.py
wardcr: Yup, good catch!

scale/recipe/models.py
wardcr: Changed to the USER trigger. In the future we would like to also add a bit more context to the event description, such that we can use it to audit what type of actions user's are requesting.

PR #500, opened 2016-09-28
Issue 54 - Batch

wardcr:
- Created batch models.
- Created system batch job type.
- Created empty batch definition.
- Added method to schedule a batch job.
- Added a test.
- Fixed formatting/warnings in job_data.


scale/batch/models.py
```diff
@@ -0,0 +1,189 @@
...
+    """
+
+    BATCH_STATUSES = (
+        ('PENDING', 'PENDING'),
+        ('QUEUED', 'QUEUED'),
+        ('RUNNING', 'RUNNING'),
+        ('FAILED', 'FAILED'),
+        ('COMPLETED', 'COMPLETED'),
+        ('CANCELED', 'CANCELED'),
+    )
```
dfaller: I would have two statuses: SUBMITTED and CREATED

SUBMITTED is when you initially put in the batch
CREATED is once the creator job is completed and the batch has been created

scale/batch/models.py
```diff
@@ -0,0 +1,189 @@
...
+        # Schedule the batch job
+        job = Queue.objects.queue_new_job(job_type, data, event)
+        batch.job = job
+        batch.save()
+
+        return batch
+
+
+class Batch(models.Model):
+    """Represents a batch of historical recipes to be re-processed on the cluster
```
dfaller: "Represents a batch of jobs and recipes to be processed on the cluster"

scale/batch/models.py
```diff
@@ -0,0 +1,299 @@
...
+        :type old_recipe: :class:`recipe.models.Recipe`
+        """
+
+        # Check whether the batch recipe already exists
+        if BatchRecipe.objects.filter(batch=batch, superseded_recipe=old_recipe).exists():
+            return
+
+        # Create the new recipe and its associated jobs
+        batch_definition = batch.get_batch_definition()
+        handler = Recipe.objects.reprocess_recipe(old_recipe.id, batch_definition.job_names, batch_definition.all_jobs)
```
wardcr: The reprocess_recipe method calls the queue method internally. Using this method instead allows us to reuse the logic for populating job names, building the recipe graph deltas, and the unit test coverage for making sure all the right jobs get created.

dfaller: Ah yes, that's right. Ok, I think everything looks good then.

Issue #344, opened 2016-06-16
Superseded Info on Job Detail Page


dfaller:
This requires the REST API changes in issue #58

Add superseded info on the job detail page:

If a job's "is_superseded" field is true, display a box (similar to the error info box) above the error info box (if both are present) that states:
Superseded: This job is obsolete because its recipe has been reprocessed.

If a job is superseded, it cannot be re-queue, so the Requeue button should not be available if the "is_superseded" field is true. This also applies to the jobs table.

We might want to pick a color to represent "superseded".

If a job's "superseded_by_job" field is not null, add a row directly under the "Recipe" row that displays:
Superseded By: *Link to the superseded_by_job that displays the job's job type and version*

If a job's "superseded_job" field is not null, add a row directly under the "Superseded By" row that displays:
Supersedes: *Link to the superseded_job that displays the job's job type and version*

The API may return multiple recipes in the "recipes" field. For the "Recipes" row, make sure you are using the latest recipe (latest created time).

PR #501, opened 2016-09-28
Issue #344 - Updated job details to display superseded info

dickmancj-mil:
Updated Job Details UI to show Superseded details. Completes #344

### DIFF
--- a/scale-ui/app/modules/jobs/controllers/jobsController.js
+++ b/scale-ui/app/modules/jobs/controllers/jobsController.js
@@ -85,7 +85,7 @@
                             '<i class="fa fa-repeat"></i></button> ' +
                         '<button ng-show="!grid.appScope.vm.readonly && !row.entity.is_superseded && row.entity.status !== \'COMPLETED\' && row.entity.status !== \'CANCELED\'" ng-click="grid.appScope.vm.cancelJob(row.entity)" class="btn btn-xs btn-default" title="Cancel Job">' +
                             '<i class="fa fa-ban"></i></button>' +
-                    '</div> {{ row.entity.status }} <span ng-if="row.entity.is_superseded" class="label label-info">SUPERSEDED</span> </div>',
+                    '</div> {{ row.entity.status }} <span type="button"  ng-if="row.entity.is_superseded" class="label label-info pull-right" tooltip-append-to-body="true" uib-tooltip="Superseded">S</span></div>',
                 filterHeaderTemplate: '<div class="ui-grid-filter-container"><select class="form-control input-sm" ng-model="grid.appScope.vm.selectedJobStatus"><option ng-selected="{{ grid.appScope.vm.jobStatusValues[$index] == grid.appScope.vm.selectedJobStatus }}" value="{{ grid.appScope.vm.jobStatusValues[$index] }}" ng-repeat="status in grid.appScope.vm.jobStatusValues track by $index">{{ status.toUpperCase() }}</option></select></div>'
             },
             {
--- a/scale-ui/app/modules/jobs/controllers/jobsController.js
+++ b/scale-ui/app/modules/jobs/controllers/jobsController.js
@@ -79,7 +79,13 @@
             {
                 field: 'status',
                 width: 150,
-                cellTemplate: '<div class="ui-grid-cell-contents"><div class="pull-right"><button ng-show="((!grid.appScope.vm.readonly) && (row.entity.status === \'FAILED\' || row.entity.status === \'CANCELED\'))" ng-click="grid.appScope.vm.requeueJobs({ job_ids: [row.entity.id] })" class="btn btn-xs btn-default" title="Requeue Job"><i class="fa fa-repeat"></i></button> <button ng-show="!grid.appScope.vm.readonly && row.entity.status !== \'COMPLETED\' && row.entity.status !== \'CANCELED\'" ng-click="grid.appScope.vm.cancelJob(row.entity)" class="btn btn-xs btn-default" title="Cancel Job"><i class="fa fa-ban"></i></button></div> {{ row.entity.status }}</div>',
+                cellTemplate: '<div class="ui-grid-cell-contents">' +
+                    '<div class="pull-right">' +
+                        '<button ng-show="((!grid.appScope.vm.readonly && !row.entity.is_superseded) && (row.entity.status === \'FAILED\' || row.entity.status === \'CANCELED\'))" ng-click="grid.appScope.vm.requeueJobs({ job_ids: [row.entity.id] })" class="btn btn-xs btn-default" title="Requeue Job">' +
+                            '<i class="fa fa-repeat"></i></button> ' +
+                        '<button ng-show="!grid.appScope.vm.readonly && !row.entity.is_superseded && row.entity.status !== \'COMPLETED\' && row.entity.status !== \'CANCELED\'" ng-click="grid.appScope.vm.cancelJob(row.entity)" class="btn btn-xs btn-default" title="Cancel Job">' +
+                            '<i class="fa fa-ban"></i></button>' +
+                    '</div> {{ row.entity.status }} <span ng-if="row.entity.is_superseded" class="label label-info">SUPERSEDED</span> </div>',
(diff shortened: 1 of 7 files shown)


dickmancj-mil:
A job that has been Superseded:
image

A job that Superseded another job:
image

dickmancj-mil:
Updated Jobs table to hide retry/cancel buttons and display SUPERSEDED badge on superseded jobs. 
image

scale-ui/app/modules/jobs/partials/jobDetailTemplate.html
```diff
@@ -27,11 +27,20 @@ <h4 class="panel-title error-title"><strong>{{ vm.job.error.category }} Error: {
...
+                            <strong>SUPERSEDED {{ vm.supersededChange }}</strong>
+                        </div>
+                    </div>
+                </div>
                 <strong>Overview</strong>
                 <table class="table table-hover table-condensed table-striped">
                     <tbody>
                         <tr><td>Type:</td><td><a ng-href="#/jobs/types/{{ vm.job.job_type.id }}"><span ng-bind-html="vm.job.job_type.getIcon()"></span> {{ vm.job.job_type.title }} {{ vm.job.job_type.version }}</a></td></tr>
                         <tr ng-show="vm.job.recipes"><td>Recipe:</td><td><a ng-href="#/recipes/recipe/{{ vm.job.recipes[0].id }}">{{ vm.job.recipes[0].recipe_type.title }} {{ vm.job.recipes[0].recipe_type.version }}</a></td></tr>
+                        <tr ng-show="vm.job.superseded_by_job.id"><td>Superseded by Job:</td><td><a ng-href="#/jobs/job/{{ vm.job.superseded_by_job.id }}">{{ vm.job.superseded_by_job.job_type.title }} {{ vm.job.superseded_by_job.job_type.version }}</a></td></tr>
```
dfaller: I would take off the "Job" and just do "Superseded by:"

It is shorter and seems cleaner.

scale-ui/app/modules/jobs/partials/jobDetailTemplate.html
```diff
@@ -27,11 +27,20 @@ <h4 class="panel-title error-title"><strong>{{ vm.job.error.category }} Error: {
                         {{ vm.job.error.description }}
                     </div>
                 </div>
+                <div class="panel panel-info" ng-show="vm.job.is_superseded">
+                    <div class="panel-heading">
+                        <div class="panel-title">
+                            <strong>SUPERSEDED {{ vm.supersededChange }}</strong>
+                        </div>
+                    </div>
+                </div>
```
dfaller: Can you add the text "This job is obsolete because its recipe has been reprocessed." to the superseded box?
I would like to communicate to the user what SUPERSEDED means.

Also I think the SUPERSEDED box might be better above the status box rather than below it. If a job is superseded, its status seems like secondary importance.

scale-ui/app/modules/jobs/controllers/jobsController.js
```diff
@@ -79,7 +79,13 @@
...
+                cellTemplate: '<div class="ui-grid-cell-contents">' +
+                    '<div class="pull-right">' +
+                        '<button ng-show="((!grid.appScope.vm.readonly && !row.entity.is_superseded) && (row.entity.status === \'FAILED\' || row.entity.status === \'CANCELED\'))" ng-click="grid.appScope.vm.requeueJobs({ job_ids: [row.entity.id] })" class="btn btn-xs btn-default" title="Requeue Job">' +
+                            '<i class="fa fa-repeat"></i></button> ' +
+                        '<button ng-show="!grid.appScope.vm.readonly && !row.entity.is_superseded && row.entity.status !== \'COMPLETED\' && row.entity.status !== \'CANCELED\'" ng-click="grid.appScope.vm.cancelJob(row.entity)" class="btn btn-xs btn-default" title="Cancel Job">' +
+                            '<i class="fa fa-ban"></i></button>' +
+                    '</div> {{ row.entity.status }} <span ng-if="row.entity.is_superseded" class="label label-info">SUPERSEDED</span> </div>',
```
dfaller: Not sure if we want to combine both the status and superseded in the same column. At the moment no superseded jobs will be displayed anyway, since they are filtered out from the REST API. We could add a checkbox or something to include superseded jobs (there is a boolean param on the REST API).

wardcr:
I don't think black is a good choice because it won't stand out to get the user's attention. I would suggest maybe orange?

scale-ui/app/modules/jobs/controllers/jobsController.js
wardcr: Agreed. I think just allowing a filter is sufficient since viewing the superseded ones in the table would be uncommon.

scale-ui/app/modules/jobs/partials/jobDetailTemplate.html
dfaller: The job part might be missing. A job can be superseded without a specific job having superseded it. So if there is no "superseded by" job, then we want just SUPERSEDED instead of SUPERSEDED BY. Agree with @wardcr suggestions.

scale-ui/app/modules/jobs/controllers/jobsController.js
dickmancj-mil: Made sense to me that after toggling future "Include Superseded" you would want to see in the list whether a job was superseded.

scale-ui/app/modules/jobs/partials/jobDetailTemplate.html
dickmancj-mil: Is this closer? Do we still want to see the Superseded By in the table below? Seems redundant. 
image

wardcr: Yeah, that's moving in the right direction. I would suggest dropping the colon and using a bit more space, maybe event right align the timestamp so the link stands out better.

dickmancj-mil: What do you think? 
image

scale-ui/app/modules/jobs/controllers/jobsController.js
dickmancj-mil: How's this? 
image
Removed Supserseded badge from Status column and added a "S" badge to JobType if job has been superseded.

scale-ui/app/modules/jobs/partials/jobDetailTemplate.html
wardcr: LGTM. Derick suggested changing the color though.

scale-ui/app/modules/jobs/controllers/jobsController.js
wardcr: I think that's better. Since it is somewhat related to "status" information and I don't believe a superseded job can have any actions it might be better to just place it where the action buttons would have been. If we leave it in it's current position it takes away more space from the title. Either way, make sure it doesn't conflict with longer titles and that it has a tool-tip.

scale-ui/app/modules/jobs/partials/jobDetailTemplate.html
dfaller: I like this a lot. Not sure if you saw my comment somewhere, but if there is no superseded by job then it should just say SUPERSEDED with the timestamp.

Yeah, I haven't been sold on any color yet. We could leave it for now and change it later. Maybe we'll get a good suggestion at review time.

scale-ui/app/modules/jobs/controllers/jobsController.js
dfaller: Yeah, I think putting it where the action buttons are makes sense.

Instead of using S, we can use an icon. I had submitted an issue a while back about job status icons and had one picked out for SUPERSEDED as well. Give f24d a try and we'll see how it looks. For a tooltip you can use the same "This job is obsolete because its recipe has been reprocessed." text.

scale-ui/app/modules/jobs/partials/jobDetailTemplate.html
dickmancj-mil: I did see that comment and it behaves just as you suggested. 
image

As far as colors, I think we need to rethink the way we display statuses. It's hard to understand the colors unless you're working with them all day every day.

scale-ui/app/modules/jobs/controllers/jobsController.js
dickmancj-mil: I just pushed this update. Moved the badge back to Status column and added tooltip. Thanks for the feedback. s
image

Issue #503, opened 2016-09-29
Eliminate UI build when unchanged to reduce time


gisjedi:
The majority of build time is spent on the retrieval of NPM packages simply for it to act as a dependency of the gulp task runner that compiles the static artifacts. This could be almost entirely eliminated if we precompiled the UI and stored a hash of all files within the UI source tree for detecting updates. Builds would only be needed on update.

dickmancj-mil:
I added support for "--tag" param on the gulp deploy task. 

Example: gulp deploy
Produces package in file deploy/scale-ui-master.tar.gz

Example: gulp deploy --tag 3.2.0
Produces package in file deploy/scale-ui-3.2.0.tar.gz

PR #505, opened 2016-09-30
#503 - Update UI to only build on change

gisjedi:
Completed:
- Moved Scale UI code to https://github.com/AppliedIS/scale-ui
- Created a Travis CI build that pushes UI artifacts to S3.
- Updated Scale Dockerfile to remove all NPM building and curl in artifacts from S3
- Reduced Scale image by another 500+ MB by making this change.

Future:
- Move appliedis/scale-ui repo to ngageoint/scale-ui
- Add support for versioned retrieval of UI artifacts marked at the same version on releases.


Dockerfile
```diff
@@ -81,25 +83,14 @@ COPY dockerfiles/framework/scale/country_data.json.bz2 /opt/scale/
...
-WORKDIR /opt/scale-ui
-RUN yum install -y npm node-gyp make \
- && npm install --global gulp-cli \
- && npm install \
- && pip install -r /tmp/docs.txt \
- && gulp deploy \
+RUN  pip install -r /tmp/docs.txt \
  && mkdir -p /opt/scale/ui \
- && tar -C /opt/scale/ui -zxf deploy/scale-ui.tar.gz \
+ && curl -L https://s3.amazonaws.com/ais-public-artifacts/scale-ui/scale-ui.tar.gz | tar -C /opt/scale/ui -zx \
```
dfaller: We need this path to not be hard-coded.

gisjedi: Agreed, I'll parameterize it.

gisjedi:
I'd argue that we should move the Scale UI into a different repo for the following reasons:

- Multiple builds are not going to be possible with Travis and it would still slow everything down ending with no real overall performance gain.
- The UI code is entirely separate of any dependency between Scale scheduler / API and so can be built, mocked and tested in isolation. I'd say at this point it gets less overall changes than we are making to Scale and all that is needed is to pull it in on build. I asked @dickmancj-mil to update the deploy step to take a tag, so it will build a tag package whenever we are ready to release.
-  It sets us up to start splitting components of the system into different Docker images / packages.

My proposal for the ultimate break-down of Scale components would be, by Docker image:

scale-web (UI from scale-ui repo and APIs / docs from scale repo)
scale-logstash (image for elasticsearch log routing from scale-logstash repo)
scale (scheduler and bootstrap of other containers from scale repo)

Dockerfile
gisjedi: Parameterized this as SCALE_UI_URL for replacement on disconnected networks.

Issue #546, opened 2016-10-19
Cleanup unneeded variables from DCOS Universe deployment


gisjedi:
We should strip out all environment variables not needed in standard deployments:

- DEPLOY_*
- NFS_*
- MESOS_MASTER_URL
- INIT_DB

All `DEPLOY_*` environment variables should be removed in favor of supporting default behaviors when the `*_ADDRESS` variables are unset. When set, the deploy step should be skipped. This will allow the default case of deployment for all pieces within a basic cluster, while still supporting any component to be externally launched and referenced.

PR #551, opened 2016-10-20
#546 - Simplify environment variables defaults for DCOS deploy

gisjedi:


Dockerfile
```diff
@@ -6,36 +6,35 @@ EXPOSE 80
...
-# ENABLE_WEBSERVER=true to start the RESTful API server, should only be set on webserver
-# DEPLOY_DB to start the database container (for DC/OS use)
-# DEPLOY_LOGGING to start up the logstash system
-# DEPLOY_WEBSERVER to start the webserver - without this we will run a naked scheduler
-# INIT_DB to initialize the database (migrate, load, etc.)
-# LOAD_COUNTRY_DATA to load country borders fixture into the database (don't select this if you have custom country data)
-# LOGSTASH_DOCKER_IMAGE the name of the DOcker image for logstash
-# SCALE_DEBUG
+# recognized environment variables
+# CONFIG_URI
```
dfaller: I think this can use a Marathon env var.

gisjedi: I looked through the documentation and I could not find a environment variables that presently exposes this inside the container. I think we should table that for now, but I'd like to go there in the future.

PR #556, opened 2016-10-20
Issue 536 - Migrate REST API to v4

wardcr:
- Changed API to require a version prefix. This uncovered an incompatibility between URL prefixes and namespaces that isn't really documented. A workaround was developed by grouping the versioned APIs into separate URL pattern groups when the master URL routes are registered. This also required updating every REST unit test to include the latest version prefix by default. It also required using a DRF method for reversing URLs instead of the standard Django one.

- Updated the logs API to support the v3/v4 behavior.

- Updated the job details and recipe details to support v3/v4 behavior.

- Updated docs for v3/v4 changes.

- Added a TODO tag "API_V3" throughout the code to make future clean up work easier.

### DIFF
--- a/scale/docs/rest/job.rst
+++ b/scale/docs/rest/job.rst
@@ -289,7 +289,7 @@ These services provide access to information about "all", "currently running" an
 | .name              | String            | The name of the input as defined by the job type interface.                    |
 |                    |                   | (See :ref:`architecture_jobs_interface_spec`)                                  |
 +--------------------+-------------------+--------------------------------------------------------------------------------+
-| .type              | String            | The type of the input as defined by teh job type interface.                    |
+| .type              | String            | The type of the input as defined by the job type interface.                    |
 |                    |                   | (See :ref:`architecture_jobs_interface_spec`)                                  |
 +--------------------+-------------------+--------------------------------------------------------------------------------+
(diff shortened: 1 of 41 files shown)


scale/docs/rest/job.rst
```diff
@@ -198,6 +198,8 @@ These services provide access to information about "all", "currently running" an
 
 .. _rest_job_details:
 
+.. TODO: API_V3 Remove attributes
```
dfaller: Still work to do here?

scale/docs/rest/job_execution.rst
```diff
@@ -353,6 +353,7 @@ These services provide access to information about "all", "currently running" an
 |  }                                                                                                                        |
 +---------------------------------------------------------------------------------------------------------------------------+
 
+.. TODO: API_V3 Remove this block
```
dfaller: Remove this line? It mentions being gone in v4.

scale/docs/rest/recipe.rst
```diff
@@ -137,6 +137,8 @@ These services provide access to information about recipes.
 
 .. _rest_recipe_details:
 
+.. TODO: API_V3 Remove attributes
```
dfaller: Do we want to document the old v3 fields?

wardcr:
Closes #556

PR #570, opened 2016-10-26
Issue 539 - Hide R&D job types

dickmancj-mil:
Added switch to Hide R&D job types to Job Types view.

image


wardcr:
I added a new REST filter parameter "is_operational" to the /job-types and /job-types/status views that can be used to get only operational, only R&D job-types, or all job types.

wardcr:
For the UI, I think it's more clear to invert the option to be something like "Show R&D" or "Include R&D".

dickmancj-mil:
What should be the default value? Show or no?

dfaller:
Let's go with show for right now.

PR #583, opened 2016-11-02
Issue 580 - Update batches to support data date

wardcr:
- Added a new RecipeFile model that connects recipes to their input files.
- Created a custom migration to back-fill the recipe file models for all existing recipes.
- Updated create recipe call to also create recipe file models for new recipes.
- Updated the batch definition to allow a data type date range.
- Updated batch system to support the new type of query.

### DIFF
--- a/scale/recipe/models.py
+++ b/scale/recipe/models.py
@@ -110,6 +110,13 @@ class RecipeManager(models.Manager):
         recipe.data = data.get_dict()
         recipe.save()
 
+        # Save models for each recipe input file
+        for input_file_id in data.get_input_file_ids():
+            recipe_file = RecipeFile()
+            recipe_file.recipe_id = recipe.id
+            recipe_file.scale_file_id = input_file_id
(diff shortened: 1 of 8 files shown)


scale/recipe/models.py
```diff
@@ -110,6 +110,13 @@ def create_recipe(self, recipe_type, data, event, superseded_recipe=None, delta=
         recipe.data = data.get_dict()
         recipe.save()
 
+        # Save models for each recipe input file
+        for input_file_id in data.get_input_file_ids():
+            recipe_file = RecipeFile()
+            recipe_file.recipe_id = recipe.id
+            recipe_file.scale_file_id = input_file_id
+            recipe_file.save()
```
dfaller: Though in practice we don't have a large number of input files for recipes yet, we should go ahead and make this a bulk create. It's possible now to create a recipe with many file inputs and in the future it will be more common.
This would require explicitly setting the created date.

PR #624, opened 2016-11-16
Issue 219 - Scale Docker clean up

dfaller:
This implements Docker cleanup for Scale job executions.

1. Job executions now have a "cluster_id" field that guarantees uniqueness across frameworks.
2. New node handling in the scheduler tracks a variety of node states. Nodes are put in a state where they must perform an initial cleanup before other tasks can be scheduled on them.
3. Created a new clean up task that gets scheduled as a Mesos command task. If performing initial clean up on a node, it deletes all non-running Docker containers and all dangling Docker volumes. If cleaning up after specific job executions, it deletes all non-running Docker containers and all Docker volumes for the given job executions.
4. Implemented a CleanupManager that handles the clean up state for each node. When job executions reach any terminal state, they get sent to the CleanupManager which creates cleanup tasks for the nodes.
5. Removed Scale Cleanup system job and associated fields from the job execution model.
6. Added a migration to increase the maximum character limits for the task_update model.


scale/job/models.py
```diff
@@ -2355,12 +2415,12 @@ class TaskUpdate(models.Model):
     """
 
     job_exe = models.ForeignKey('job.JobExecution', on_delete=models.PROTECT)
-    task_id = models.CharField(max_length=50)
-    status = models.CharField(max_length=50)
+    task_id = models.CharField(max_length=250)
```
dfaller: I think having a lax length is good in case something really goes sideways; my initial restrictions were just way too strict. I think max 250 is appropriate for these fields (if the length is over 250 for any of these, something IS really wrong and we should see it as an error).

scale/job/execution/running/tasks/job_task.py
```diff
@@ -1,12 +1,12 @@
...
 from __future__ import unicode_literals
 
-from job.execution.running.tasks.base_task import Task
+from job.execution.running.tasks.exe_task import JobExecutionTask
 from job.resources import NodeResources
 
 
-class JobTask(Task):
-    """Represents a job execution job task (runs the actual job/algorithm)
+class JobTask(JobExecutionTask):
```
dfaller: I agree it's confusing. The job execution is split into three tasks: pre-task, job-task, and post-task. The job-task is confusingly named, but I didn't have a better alternative at the time. Maybe main-task? I am definitely open to better suggestions.

scale/job/configuration/configuration/job_configuration.py
```diff
@@ -288,17 +293,18 @@ def configure_logging_docker_params(self, job_exe_id):
             es_urls = ','.join(hosts)
 
         if settings.LOGGING_ADDRESS is not None:
-            self.add_pre_task_docker_param(DockerParam('log-driver', 'syslog'))
-            self.add_pre_task_docker_param(DockerParam('log-opt', 'syslog-address=%s' % settings.LOGGING_ADDRESS))
-            self.add_pre_task_docker_param(DockerParam('log-opt', 'tag=scale_%d_pre' % job_exe_id))
+            if not job_exe.is_system:
+                self.add_pre_task_docker_param(DockerParam('log-driver', 'syslog'))
```
dfaller: I changed them to take a list of Docker params and updated all of the callers. It did shorten some of the code a decent bit.

scale/job/models.py
```diff
@@ -1309,10 +1307,29 @@ def configure_docker_params(self, framework_id, workspaces):
...
+        (is no longer queued).
+
+        :returns: The cluster ID for the job execution
+        :rtype: string
+
+        :raises Exception: If the job execution is still queued
+        """
+
+        if self.status == 'QUEUED':
+            raise Exception('cluster_id is not set until the job execution is scheduled')
```
dfaller: I made a custom exception for critical Scale logic bugs and used it in all of the places you mentioned.

scale/job/models.py
```diff
@@ -1495,6 +1477,96 @@ def get_log_text(self, include_stdout=True, include_stderr=True, since=None, htm
...
+        scheduled (is no longer queued) and if this is not a system job type.
+
+        :returns: The post-task ID for the job execution
+        :rtype: string
+
+        :raises Exception: If the job execution is still queued or is a system job type
+        """
+
+        if self.is_system:
+            raise Exception('System jobs do not have a post-task')
```
dfaller: Switched to new ScaleLogicBug exception

scale/job/models.py
```diff
@@ -1495,6 +1477,96 @@ def get_log_text(self, include_stdout=True, include_stderr=True, since=None, htm
...
+        scheduled (is no longer queued) and if this is not a system job type.
+
+        :returns: The pre-task ID for the job execution
+        :rtype: string
+
+        :raises Exception: If the job execution is still queued or is a system job type
+        """
+
+        if self.is_system:
+            raise Exception('System jobs do not have a pre-task')
```
dfaller: Switched to new ScaleLogicBug exception

scale/job/execution/running/tasks/base_task.py
```diff
@@ -248,49 +191,66 @@ def get_resources(self):
...
-        :param job_exe: The job execution model
-        :type job_exe: :class:`job.models.JobExecution`
+        :param when: The time that the task was scheduled
+        :type when: :class:`datetime.datetime`
         """
 
-        pass
+        with self._lock:
+            if self._has_started:
+                raise Exception('Trying to schedule a task that has already started')
```
dfaller: Switched to new ScaleLogicBug exception

scale/job/execution/running/tasks/update.py
```diff
@@ -0,0 +1,95 @@
...
+        """
+
+        if not task_id:
+            raise Exception('task_id must be provided')
+        if not agent_id:
+            raise Exception('agent_id must be provided')
+        if status not in self.VALID_STATUSES:
+            raise Exception('invalid status')
+        if not when:
+            raise Exception('when must be provided')
```
dfaller: Switched to new ScaleLogicBug exception

dfaller:
1. The properties are to signify/allow public read, but not public write. There's a lot of multi-threaded stuff going on, and I wanted clear boundaries to assist developers. I don't think adding the properties is too much of a burden.

2. There are a lot of manager classes involved in the scheduler that need to be passed to the threads. The number of parameters is too large, so I created a container class to pass in the many managers as one parameter instead of going a global/singleton route. If there is a more Pythonic way to address this, I'm open to suggestions.

scale/job/execution/running/tasks/job_task.py
wardcr: Yeah, I think just about any name that doesn't reuse an existing concept for a different purpose is better in this case. Something like main would be fine.

wardcr:
1. Using properties for read-only purposes is a valid use case. Not saying this applies to all of the examples here, but as a general rule I just don't like it when code docs devolve into useless boilerplate such as property "name" - "Gets the name" or whatever.

2. I would say in general, by using imports based on examples from existing 3rd party projects like say the Django models API, rather than constructor arguments. I haven't studied the bigger picture of all this threading code to say for sure if that would make sense in this context though.

Issue #633, opened 2016-11-22
Logs /combined API returns more than execution specified


gisjedi:
The recent change to logging to include the full framework ID exposed issues with the Elasticsearch queries being issued. We were using queries against the analyzed string fields `scale_job_exe` and `stream` instead of explicit `term` queries against the `raw` subfields.  The tokenized values in `scale_job_exe` meant that we were retrieving logs for _all_ job executions associated with the current framework, i.e. all job executions.

PR #635, opened 2016-11-22
#633: Fix for matching all logs 

gisjedi:
Fix to use exact (not-analyzed) raw strings for execution ID and output type for log retrieval


gisjedi:
Closes #633

PR #639, opened 2016-11-28
Issue 573

mikenholt:
Summary of changes:
 * Moved `DockerParam` and `TaskWorkspace` from `job_configuration.py` to `job_parameter.py`
 * Updated `job_configuration` schema to 1.1, converted `job_configuration.py` to `job_configuration_1_0.py`
 * Updated `job_interface` schema to 1.2, converted `job_interface.py` to `job_interface_1_1.py`
 * Modified `/job/models.py` to add defined `env_vars` to job task docker params and populate the command_arguments with `settings`.


scale/job/configuration/configuration/job_configuration.py
```diff
@@ -137,332 +124,143 @@ def __init__(self, configuration=None):
...
+                raise InvalidJobConfiguration('Duplicate setting %s in job task' % name)
             self._job_task_workspace_names.add(name)
-        for workspace_dict in self._configuration['post_task']['workspaces']:
-            name = workspace_dict['name']
+        for setting_dict in self._configuration['post_task']['settings']:
+            name = setting_dict['name']
             if name in self._post_task_workspace_names:
-                raise InvalidJobConfiguration('Duplicate workspace %s in post task' % name)
+                raise InvalidJobConfiguration('Duplicate setting %s in post task' % name)
             self._post_task_workspace_names.add(name)
```
dfaller: Need to check each task to see if any settings are duplicated. Also please fix the docstring (mentions workspace names)

scale/job/configuration/interface/job_interface_1_0.py
```diff
@@ -443,6 +443,17 @@ def _check_param_name_uniqueness(self):
...
+            for setting in self.definition['settings']:
+                if setting['name'] in self._param_names:
+                    raise InvalidInterfaceDefinition('Setting names must be unique')
+                self._param_names.add(setting['name'])
+
+            for env_var in self.definition['env_vars']:
+                if env_var['name'] in self._param_names:
+                    raise InvalidInterfaceDefinition('Enviorment variable names must be unique')
+                self._param_names.add(env_var['name'])
+
```
dfaller: This code should be in the 1.2 schema, not 1.0.

While settings do have to be unique against all of the other params, the env var names only have to be unique compared to each other.

scale/job/models.py
```diff
@@ -973,6 +974,9 @@ def queue_job_exes(self, jobs, when):
             interface = job.get_job_interface()
             data = job.get_job_data()
             job_exe.command_arguments = interface.populate_command_argument_properties(data)
+            # Add configuration values for the settings to the command line.
+            job_exe.command_arguments = interface.populate_command_argument_settings(job_exe.command_arguments,
+                                                                                     job.get_job_configuration())
```
dfaller: This needs to be done at schedule time, not queue time.

scale/job/models.py
```diff
@@ -1302,6 +1306,13 @@ def configure_docker_params(self, workspaces, docker_volumes):
         # Configure any Docker parameters needed for workspaces
         configuration.configure_workspace_docker_params(self, workspaces, docker_volumes)
 
+        # Add job environment variable as docker parameters
+        interface = self.get_job_interface().get_dict()
+        for env_var in interface['env_vars']:
+            env_var_name = env_var['name']
+            env_var_value = env_var['value']
+            configuration.add_job_task_docker_params([DockerParam('env', env_var_name + '=' + env_var_value)])
+
```
dfaller: The env var values need to have the same string replacement done to them that is done to the command arguments.

scale/job/models.py
```diff
@@ -2420,4 +2431,4 @@ class TaskUpdate(models.Model):
 
     class Meta(object):
         """Meta information for the database"""
-        db_table = 'task_update'
+        db_table = 'task_update'
```
dfaller: Add missing newline.

scale/job/configuration/interface/job_interface.py
```diff
@@ -280,3 +282,42 @@ def populate_command_argument_settings(self, command_arguments, job_configuratio
...
+        """
+
+        for env_var in self.definition['env_vars']:
+            env_var_name = env_var['name']
+            env_var_value = env_var['value']
+            command_arguments = self._replace_command_parameter(command_arguments,
+                                                                env_var_name,
+                                                                env_var_value)
+
+        return command_arguments
```
dfaller: The env vars are supposed to be substituted into the command arguments. The same substitution done into the command arguments is supposed to be done into the env vars.

So a command argument can have input params, settings, and the job output dir replaced in its string, e.g. ${param1} ${setting1} ${job_output_dir}.

Env vars may have the same substitution performed with their values e.g. "env_vars": [{"name": "var1", "value": "${param1}"].

Also the env vars need to be passed from the job interface into the Docker params in the job_exe model's job_configuration. So at schedule time, pull env vars from interface, perform any substitution on their values, and insert them into the job_configuration Docker params.

scale/job/models.py
```diff
@@ -2431,4 +2436,5 @@ class TaskUpdate(models.Model):
 
     class Meta(object):
         """Meta information for the database"""
-        db_table = 'task_update'
\ No newline at end of file
+
+        db_table = 'task_update'
```
dfaller: It looks like an extra line was added here between the docstring and db_table.

Issue #615, opened 2016-11-14
S3 workspace and Strike configuration bug


dfaller:
The S3 workspace and Strike configurations should ignore access key credentials and host paths that are set to empty strings. Right now the UI is inserting empty strings into those fields and it is causing the S3 strike to fail.

gisjedi:
For the issue of empty keys for credentials, the current behavior actually is that empty values raise an exception, but empty 1+ length strings do not. I propose that we change this to be whitespace pruned 0 length strings quietly fall-back to role-based auth. Actually missing `access_key_id` or `secret_access_key` should raise an exception as that is a truly invalid `credentials` object.

I'm going to argue that we should fix the UI to not be passing these empty string values as I would prefer are API be able to assume rational values for the `credentials` object, as opposed to having to assume empty values means we should fallback to role-based auth. I'm going to propose a new issue to address this sort of behavior throughout the UI.

PR #640, opened 2016-11-28
#615 - S3 strike and broker configuration hardening

gisjedi:
Handling for a number of whitespace filled values provided via UI. host_path, credentials.access_key_id and credentials.secret_access_key now all will probably be handled when only containing whitespace.

### DIFF
--- a/scale/util/aws.py
+++ b/scale/util/aws.py
@@ -62,9 +62,12 @@ class AWSClient(object):
     def instantiate_credentials_from_config(config):
         """Extract credential keys from configuration and return instantiated credential object
 
+        If values provided in the `access_key_id` or `secret_access_key` keys of `credentials` configuration object are
+        empty, None will be returned. In this case, role-based authentication should be attempted.
+
         :param config: Resource specific configuration
         :type config: :class:`botocore.client.Config`
-        :return: instantiated credential object
+        :return: instantiated credential object or None if keys provided were empty
         :rtype: :class:`util.aws.AWSCredentials`
 
         :raises :class:`util.exceptions.InvalidAWSCredentials`: If the credentials provided are incomplete.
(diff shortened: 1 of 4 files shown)


scale/util/aws.py
```diff
@@ -60,13 +60,31 @@ def __exit__(self, type, value, traceback):
...
                 raise InvalidAWSCredentials('"credentials" requires "secret_access_key" to be populated')
-            return AWSCredentials(credentials_dict['access_key_id'], credentials_dict['secret_access_key'])
+
+            access_key = credentials_dict['access_key_id'].strip()
+            secret_key = credentials_dict['secret_access_key'].strip()
+
+            # If either Access Key or Secret Access Key are empty, fail-over to role-based auth.
+            # TODO: This should be removed once the UI has been improved to prune unset values from requests.
+            if not len(access_key) or not len(secret_key):
+                return None
```
dfaller: If the method is going to return None now, the method documentation (and possibly calling code) should be updated to reflect that.

scale/util/aws.py
```diff
@@ -60,13 +60,31 @@ def __exit__(self, type, value, traceback):
...
-            if 'secret_access_key' not in credentials_dict or not credentials_dict['secret_access_key']:
+            if 'secret_access_key' not in credentials_dict:
                 raise InvalidAWSCredentials('"credentials" requires "secret_access_key" to be populated')
-            return AWSCredentials(credentials_dict['access_key_id'], credentials_dict['secret_access_key'])
+
+            access_key = credentials_dict['access_key_id'].strip()
+            secret_key = credentials_dict['secret_access_key'].strip()
+
+            # If either Access Key or Secret Access Key are empty, fail-over to role-based auth.
+            # TODO: This should be removed once the UI has been improved to prune unset values from requests.
```
dfaller: After the UI is fixed, we should still check the fields for blank values and raise InvalidAWSCredentials since a blank value would never be correct.

PR #646, opened 2016-11-30
Issue 629 - Improve scheduler task handling

dfaller:
Scale now restarts lost tasks without failing the job execution.
The "Lost by Mesos" error has been removed.
Reconciliation is now performed on job execution tasks as well.
Changed the way scheduler managers are accessed.
Switched cleanup tasks to use framework ID in their task IDs instead of agent ID


dfaller:
There was a test for lost tasks. I have added a test for task reconciliation.

Issue #650, opened 2016-12-06
Allow users to hook analytics up to the scale interface


bowmanmc:
Allow integration of organizations analytics tools (google analytics, piwik, etc...) in the scale user interface.

PR #652, opened 2016-12-06
Fixes issue #650 - add analytics support

bowmanmc:
injects the contents of config/analytics.html into the index.html
document as part of the user interface build process.


gisjedi:
@bowmanmc Just make sure the tars for the dependencies are updated.  Ref the document here to do so: https://github.com/ngageoint/scale/blob/master/scale-ui/README.md#dependencies

bowmanmc:
Ok, added gulp-inject to node_modules.tar.gz and double checked there weren't any child deps missing.

gisjedi:
This LGTM and since Michael isn't an org member, I'm going to merge.

PR #672, opened 2016-12-15
Issue 665 - Default job type settings

mikenholt:
This allows for default settings of a job_type to be defined.  This branch also includes a couple changes I made to clean up several files.

### DIFF
--- a/scale/job/configuration/interface/configuration_interface.py
+++ b/scale/job/configuration/interface/configuration_interface.py
@@ -34,7 +34,7 @@ CONFIGURATION_INTERFACE_SCHEMA = {
 class ConfigurationInterface(object):
     """Represents the interface for defining a default job configuration"""
 
-    def __init__(self, definition):
+    def __init__(self, definition=None):
         """Creates a configuration interface from the given definition.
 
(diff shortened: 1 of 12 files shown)


scale/job/models.py
```diff
@@ -1030,6 +1031,10 @@ def schedule_job_executions(self, framework_id, job_executions, workspaces):
             if resources is None:
                 raise Exception('Cannot schedule job execution %i without resources' % job_exe.id)
 
+            # Populate job_configuration with default settings
+            configuration = job_exe.get_job_configuration()
+            configuration.populate_default_job_settings(job_exe)
```
dfaller: This doesn't get set back to the job_exe.configuration field. Move this to configure_docker_params(), somewhere near the top is fine.

scale/job/configuration/interface/configuration_interface.py
```diff
@@ -0,0 +1,96 @@
...
+            'item': {
+                'type': 'string',
+            },
+        },
+    },
+}
+
+
+class ConfigurationInterface(object):
+    """Represents the interface for defining a default job configuration"""
```
dfaller: I would rename this to JobTypeConfiguration (and change all applicable code with the configuration interface name).

PR #680, opened 2016-12-21
Fixed settings population and required settings check.

mikenholt:
Simplified the required settings check and added a call to populate the command_args in the pre-task.

### DIFF
--- a/scale/job/configuration/interface/job_interface.py
+++ b/scale/job/configuration/interface/job_interface.py
@@ -366,7 +366,7 @@ class JobInterface(previous_interface.JobInterface):
             raise InvalidInterfaceDefinition('Environment variable names must be unique')
 
     def validate_populated_settings(self, job_exe, job_configuration):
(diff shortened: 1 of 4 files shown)


scale/job/configuration/interface/job_interface.py
```diff
@@ -375,24 +375,14 @@ def validate_populated_settings(self, job_exe, job_configuration):
         """
 
         interface_settings = self.definition['settings']
-        configuration_settings = {setting.name: setting.value for setting in job_configuration.get_job_task_settings()}
 
-        command_arguments = job_exe.command_arguments
-        env_vars = self.populate_env_vars_arguments(job_configuration)
-        env_var_str = ' '.join(env_var['value'] for env_var in env_vars)
+        configuration_settings = {setting.name: setting.value for setting in job_configuration.get_job_task_settings()}
+        config_setting_names = list(configuration_settings.keys())
```
dfaller: This could be simplified into one line.

PR #691, opened 2017-01-04
S3 bucket validation should catch credential errors

mikenholt:
Added additional except case to catch S3 credentials that are missing from the workspace.


scale/storage/brokers/s3_broker.py
```diff
@@ -121,6 +121,9 @@ def validate_configuration(self, config):
             except ClientError:
                 warnings.append(ValidationWarning('bucket_access',
                                                   'Unable to access bucket. Check the bucket name and credentials.'))
+            except NoCredentialsError:
```
dfaller: You should be able to combine this in the same line as catching the ClientError.

PR #709, opened 2017-01-30
Issue 708 - Scale file refactor

dfaller:
This PR refactors the scale_file table to improve query performance related to source and product files.

Through a series of migrations, the source_file and product_file fields are added to the scale_file table, which will now contain all file fields. A new file_type field has been added as well to distinguish between 'SOURCE' and 'PRODUCT' files. Moving everything to one table improves the ability to index the data appropriately and eliminates the joins previously needed to the source_file and product_file tables.

The SourceFile and ProductFile models are now Django proxy models, which basically have no affect on the database tables and offer model functionality specific to source/product files through their managers.

All Scale manager logic has been updated to reflect the model changes.

### DIFF
--- a/scale/cli/management/commands/migratedata.py
+++ b/scale/cli/management/commands/migratedata.py
@@ -105,7 +105,7 @@ class Command(BaseCommand):
                 logging.info("Processing ingest %s" % ingest.file_name)
                 with transaction.atomic():
                     ingest.ingest_started = datetime.utcnow()
-                    sf = ingest.source_file = SourceFile()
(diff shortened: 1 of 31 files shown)


dfaller:
I didn't see any way to affect the SourceFile.objects.create() methods, but they only appear a couple times in the test utilities. I created a static create() method on SourceFile and ProductFile to set file_type and those are used by the actual Scale logic. At any rate, very little of the effort involved setting file_type as it was only in a few places.

PR #712, opened 2017-02-02
Issue 700 - Errors for missing input files

dfaller:
This PR revolved around adding system errors for missing and deleted input files in the pre-task.

1. Refactored how system errors and exit codes are handled for pre/post tasks. The newer system should be easier to maintain.
2. Added new errors for deleted and missing files that are checked in the pre-task.
3. Added a new flag to the error model that indicates if the error should be automatically retried. The SYSTEM category is no longer used for this.

### DIFF
--- a/scale/error/exceptions.py
+++ b/scale/error/exceptions.py
@@ -2,7 +2,7 @@
 from __future__ import unicode_literals
 
 import logging
-from abc import ABCMeta
+from abc import ABCMeta, abstractmethod
 
 from django.db.utils import DatabaseError, OperationalError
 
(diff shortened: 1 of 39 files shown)


scale/error/exceptions.py
```diff
@@ -0,0 +1,157 @@
...
+
+    def get_log_message(self):
+        """The log message to print out when the error occurs. This should be overridden by derived classes to provide
+        detailed log messaging.
+
+        :returns: The log message
+        :rtype: string
+        """
+
+        return 'An error occurred. Please override this log message to provide more detail.'
```
gisjedi: Is there a case where this does get called without being overridden? If not just `raise NotImplementedError`

scale/error/exceptions.py
```diff
@@ -2,7 +2,7 @@
 from __future__ import unicode_literals
 
 import logging
-from abc import ABCMeta
+from abc import ABCMeta, abstractmethod
```
gisjedi: Sweet, I wasn't aware of the abstractmethod annotation.

PR #725, opened 2017-02-17
Issue 669 - Search source files by data time

dfaller:
Added a new time_field to the /sources/ and /sources/updates/ APIs to allow the user to filter by data time instead of last modified time. I also tweaked the Scale file indexes by removing the index for file_type and creating multi-column indexes for (file_type, data_started, data_ended) and (file_type, last_modified).


scale/storage/migrations/0007_auto_20170217_0901.py
```diff
@@ -0,0 +1,24 @@
...
+    operations = [
+        migrations.AlterField(
+            model_name='scalefile',
+            name='file_type',
+            field=models.CharField(default='SOURCE', max_length=50, choices=[('SOURCE', 'SOURCE'), ('PRODUCT', 'PRODUCT')]),
+            preserve_default=True,
+        ),
+        migrations.AlterIndexTogether(
+            name='scalefile',
+            index_together=set([('file_type', 'last_modified'), ('file_type', 'data_started', 'data_ended')]),
```
gisjedi: It may be worth investigating whether your multi-column index is optimal. I'd ref this Stack Overflow post answer by Erwin Brandstetter on ordering of multiple time values in your index: http://dba.stackexchange.com/questions/39589/optimizing-queries-on-a-range-of-timestamps-two-columns

dfaller: I removed the multi-column index. Based on some testing, I think the existing single column indexes are sufficient for now. Changing the index order (ASC, DESC) on a multi-column index is not supported on our version of Django. We need to upgrade as it was recently implemented.

PR #731, opened 2017-02-22
Issue 721

tclarke:
Adds minio/nats workspace and strike support for an S3-like experience in your very own home.
Includes:

- backend/documentation changes to allow the S3 broker to work with minio (specifically, allows specification of a server URL)
- a new monitor type for one of the supported minio notification systems (nats streaming server) which is fairly simple and fast
- GUI changes to support the broker modifications
- GUI change to allow the display of nats monitor properties in the strike details view
- A dockerfile which creates a basic minio and NATS streaming server image which can be used with the new monitor type.

Does not include GUI changes to allow the creation or proper viewing of nats based strike processes. Currently the GUI only allows 1-to-1 broker to monitor mappings so it will force the creation of SQS based monitoring with an S3 workspace. This change is a little more involved so I'll create a new issue to address this. The scale RESTful API can be used directly (or via the scale cli) to create appropriate strike processes.


tclarke:
Failed build fixed in #732

gisjedi:
Closing as its was an earlier version of #737.

Issue #589, opened 2016-11-07
Ingest Scan


dfaller:
Develop a capability for the ingest process to "scan" a workspace already filled with existing files.

wardcr:
This looks like a duplicate of #575

dfaller:
Yeah, though the other issue also talks about ingesting via REST API. When it comes time to implement, we'll grab this or the other item.

gisjedi:
I'm taking this and beginning the investigation on what all is needed to do a full / filtered workspace scan.

PR #744, opened 2017-02-28
#589 - Workspace Scanning Support

gisjedi:
Added support for scanning workspaces backed by both HOST and S3 type brokers. The entry point for this is through the `list_files` method on the `Workspace` model. This returns the file name and size as a `FileDetails` namedtuple delivered either via callback or on return. The `Scanner` classes originate calls to the `list_files` method and are responsible for processing the results returned by `list_files`. Scanners will queue ingests as batches are returned from the callback which will then be picked up by the scheduler. Scan configurations are required to control the parameters that a Scanner applies. These configurations mirror a Strike configuration almost identically, save for the addition of a `recursive` parameter that allows for full traversal of a multi-level workspace.


scale/ingest/scan/scanners/scanner.py
```diff
@@ -0,0 +1,317 @@
...
+
+    @abstractmethod
+    def load_configuration(self, configuration):
+        """Loads the given configuration
+
+        :param configuration: The configuration as a dictionary
+        :type configuration: dict
+        """
+
+        raise NotImplementedError
```
dfaller: Should it be `raise NotImplementedError()` ?

scale/ingest/scan/scanners/scanner.py
```diff
@@ -0,0 +1,317 @@
...
+        :param configuration: The configuration as a dictionary
+        :type configuration: dict
+        :returns: A list of warnings discovered during validation
+        :rtype: [:class:`ingest.scan.configuration.scan_configuration.ValidationWarning`]
+
+        :raises :class:`ingest.scan.scanners.exceptions.InvalidScannerConfiguration`: If the given configuration is
+            invalid
+        """
+
+        raise NotImplementedError
```
dfaller: Should it be `raise NotImplementedError()` ?

scale/ingest/scan/scanners/scanner.py
```diff
@@ -0,0 +1,317 @@
...
+                logging.info('Removed duplicate file_name %s from ingests at file_path %s',
+                             ingest.file_name, ingest.file_path)
+
+        final_ingests = [x for x in deduplicated_ingests if x.file_name not in existing_ingest_file_names]
+
+        logger.info('Removed %i duplicates of pre-existing ingests.', list_count - len(final_ingests))
+
+        return final_ingests
+
+    def _process_ingest(self, file_path, file_size):
```
dfaller: This seems like it should be generic to both scan and strike. There should be one code path for creating/setting up an ingest record and filling out the fields based on file ingest rules.

scale/ingest/scan/scanners/scanner.py
```diff
@@ -0,0 +1,317 @@
...
+                logger.info('Rule match, %s will be registered as %s on workspace %s', file_name, file_path,
+                            workspace_name)
+        else:
+            logger.info('No rule match for %s, file is being skipped', file_name)
+            ingest = None
+
+        return ingest
+
+    @transaction.atomic
+    def _start_ingest_tasks(self, ingests):
```
dfaller: This should also be generic to both scan and strike.

scale/storage/brokers/broker.py
```diff
@@ -71,7 +71,7 @@ def delete_files(self, volume_path, files):
         :type files: [:class:`storage.models.ScaleFile`]
         """
 
-        pass
+        raise NotImplementedError
```
dfaller: I think it should be `raise NotImplementedError()` here and everywhere else in the file.

scale/storage/brokers/host_broker.py
```diff
@@ -62,6 +61,62 @@ def get_file_system_paths(self, volume_path, files):
...
+
+        # Handle a full recursive walk of the directory tree.
+        if recursive:
+            for root, dirs, files in os.walk(volume_path):
+                for name in files:
+                    file_name = os.path.join(root, name)
+                    if os.path.isfile(file_name):
+                        file_details = FileDetails(file_name, os.path.getsize(file_name))
+                        file_batch.append(file_details)
+                        if len(file_batch) % batch_size == 0:
```
dfaller: `if len(file_batch) >= batch_size` seems a little clearer (easier to read) to me, but it's not a big deal.

scale/storage/brokers/host_broker.py
```diff
@@ -62,6 +61,62 @@ def get_file_system_paths(self, volume_path, files):
...
+                            try:
+                                if callback:
+                                    callback(file_batch)
+                                    file_batch = []
+                            except:
+                                logger.exception('list_files callback failure.')
+                            file_list.extend(file_batch)
+        # Handle identifying files only from a single directory.
+        else:
+            for result in os.listdir(volume_path):
```
dfaller: This has a lot of duplicate code from above. Recommend factoring that out into its own method.

scale/storage/test/utils.py
```diff
@@ -54,6 +55,22 @@ def create_file(file_name='my_test_file.txt', media_type='text/plain', file_size
         scale_file.countries = countries
         scale_file.save()
     return scale_file
+    
+    
+def create_uningested_file(file_name='my_test_file.txt', file_path=None, contents=None):
+    """Creates a raw file for use in testing list_files broker
```
dfaller: Is creating an actual file necessary for testing? Might be better to just mock any file IO calls that would interact with the file.

scale/ingest/scan/scanners/scanner.py
gisjedi: BLUF: Both class and instance of are valid forms.
TLDR -->
See PEP 3109 as a reference: `raise EXCEPTION is used to raise a new exception. This form has two sub-variants: EXCEPTION may be an exception class or an instance of an exception class; valid exception classes are BaseException and its subclasses [5]. If EXCEPTION is a subclass, it will be called with no arguments to obtain an exception instance.`

scale/ingest/scan/scanners/scanner.py
gisjedi: Same as above

scale/storage/brokers/broker.py
gisjedi: Same as above

scale/ingest/models.py
```diff
@@ -64,6 +68,35 @@ def __init__(self, strike, most_recent=None, files=0, size=0, values=None):
...
+        if scan_id:
+            ingest.scan_id = scan_id
+        if strike_id:
+            ingest.strike_id = strike_id
+
+        ingest.file_name = file_name
+        ingest.media_type = get_media_type(ingest.file_name)
+        ingest.workspace = workspace
+
+        logger.info('New file on %s: %s', ingest.workspace.name, ingest.file_name)
```
dfaller: I would recommend taking out the log statement here, as it seems a little low level. Logging in strike and scan would have more context.

PR #750, opened 2017-03-07
Issue 740 - Source Jobs REST API

dfaller:
This PR adds a REST API call to retrieve the jobs associated with a source file. It includes all standard filtering query parameters available on the normal /jobs/ API.

URL: /sources/{source-id}/jobs/

### DIFF
--- a/scale/source/models.py
+++ b/scale/source/models.py
@@ -8,6 +8,7 @@ from django.db import transaction
 from django.utils.timezone import now
 
 import storage.geospatial_utils as geo_utils
+from job.models import Job
 from source.triggers.parse_trigger_handler import ParseTriggerHandler
 from storage.brokers.broker import FileMove
 from storage.models import ScaleFile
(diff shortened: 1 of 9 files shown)


scale/source/models.py
```diff
@@ -33,6 +34,52 @@ def get_source_file_by_name(self, file_name):
...
+        :param include_superseded: Whether to include jobs that are superseded.
+        :type include_superseded: bool
+        :param order: A list of fields to control the sort order.
+        :type order: [string]
+        :returns: The list of jobs that match the time range.
+        :rtype: [:class:`job.models.Job`]
+        """
+
+        # Order must end with job ID so we can do a distinct on the job ID and remove duplicate jobs
+        if order:
```
wardcr: It might be good to add a unit test or two to cover these special case manipulations of the order parameter, although you do have one in the view tests, so not super urgent.

scale/source/models.py
```diff
@@ -8,6 +8,7 @@
 from django.utils.timezone import now
 
 import storage.geospatial_utils as geo_utils
+from job.models import Job
```
wardcr: I wasn't sure if it's okay for the source app to depend on job or if it should be inlined?

scale/source/urls.py
```diff
@@ -9,5 +9,6 @@
     url(r'^sources/updates/$', views.SourceUpdatesView.as_view(), name='source_updates_view'),
     url(r'^sources/(?P<source_id>\d+)/$', views.SourceDetailsView.as_view(), name='source_details_view'),
     url(r'^sources/(?P<file_name>[\w.]{0,250})/$', views.SourceDetailsView.as_view(), name='source_details_view'),
+    url(r'^sources/(?P<source_id>\d+)/jobs/$', views.SourceJobsView.as_view(), name='source_jobs_view'),
```
wardcr: Since the sources details view supports an ID or file name (requested by users), it would be good if any of the sub-resources also supported file name for consistency.

scale/source/models.py
dfaller: Yes, source can rely on job.

scale/source/urls.py
dfaller: I think we're going away from the concept of unique source file names, so I didn't include that at the moment.

PR #752, opened 2017-03-08
Issue 585 - Batch trigger files for recipes

wardcr:
- Updated batch definition to support trigger rules for existing files using either the recipe types's default rule or a completely custom rule.
- Trigger rule implementation is modeled after the existing ingest triggers.
- Updated batch processor, REST views create/validate, tests, and docs to support the new definition.


scale/batch/configuration/definition/batch_definition.py
```diff
@@ -63,10 +69,65 @@
                 },
             },
         },
+        'trigger_rule_item': {
+            'description': 'A range of dates used to determine which recipes should be re-processed',
```
dfaller: Correct the description

Issue #742, opened 2017-02-27
API needed for launching workspace scan


gisjedi:
Workspace scanning capability is almost complete and an associated REST API endpoint is needed to expose.

gisjedi:
This is dependent on #589

PR #756, opened 2017-03-10
#742 - Workspace Scan API

gisjedi:
Added ability to construct Workspace Scan models and kick off dry run and full ingest process from them via REST API. Documentation is also added for Scan to the REST and Architecture sections.


scale/ingest/models.py
```diff
@@ -723,7 +721,44 @@ def get_details(self, scan_id):
...
+        atomic transaction.
+
+        :param scan_id: The unique identifier of the Scan process.
+        :type scan_id: int
+        :param dry_run: Whether the scan will execute as a dry run
+        :type dry_run: bool
+        :returns: The new Scan process
+        :rtype: :class:`ingest.models.Scan`
+        """
+
```
dfaller: We need to think about overwriting an old dry or normal job. We probably don't want to re-run a normal job, not sure about dry runs.

scale/ingest/urls.py
```diff
@@ -12,6 +12,13 @@
     url(r'^ingests/(?P<ingest_id>\d+)/$', views.IngestDetailsView.as_view(), name='ingest_details_view'),
     url(r'^ingests/(?P<file_name>[\w.]{0,250})/$', views.IngestDetailsView.as_view(), name='ingest_details_view'),
 
+    # Scan views
+    url(r'^scans/$', views.ScansView.as_view(), name='scans_view'),
+    url(r'^scans/$', views.ScansView.as_view(), name='scans_view'),
```
dfaller: Duplicate line

scale/ingest/views.py
```diff
@@ -42,10 +44,16 @@ def list(self, request):
...
         file_name = rest_util.parse_string(request, 'file_name', required=False)
         order = rest_util.parse_string_list(request, 'order', required=False)
 
-        ingests = Ingest.objects.get_ingests(started, ended, ingest_statuses, strike_ids, file_name, order)
+        ingests = Ingest.objects.get_ingests(started=started, ended=ended,
+                                             statuses=ingest_statuses,
+                                             scan_ids=scan_ids,
+                                             strike_ids=strike_ids,
+                                             file_name=file_name,
+                                             order=order)
```
dfaller: Need to update ingest REST API docs with the new parameter.

gisjedi:
During dry runs and prior to an actual ingest, configuration changes should be allowed. Once you have launched the ingest job, no further dry runs or configuration changes will be allowed. We will continue to allow multiple dry runs and only the newest dry run job will be related to the Scan model.

scale/docs/rest/scan.rst
```diff
@@ -0,0 +1,535 @@
...
+| name               | String            | Required | The identifying name of the Scan process used for queries.          |
++--------------------+-------------------+----------+---------------------------------------------------------------------+
+| title              | String            | Optional | The human readable display name of the Scan process.                |
++--------------------+-------------------+----------+---------------------------------------------------------------------+
+| description        | String            | Optional | A longer description of the Scan process.                           |
++--------------------+-------------------+----------+---------------------------------------------------------------------+
+| configuration      | JSON Object       | Required | JSON defining the Scan configuration.                               |
+|                    |                   |          | (See :ref:`architecture_scan_spec`)                                 |
++--------------------+-------------------+----------+---------------------------------------------------------------------+
```
gisjedi: Nope, as that will be identified in the `ingest` parameter given to the `scan/{id}/process` API call.

scale/ingest/models.py
```diff
@@ -626,15 +632,6 @@ def create_scan(self, name, title, description, configuration, dry_run=True):
...
-        scan_type = self.get_scan_job_type()
-        job_data = JobData()
-        job_data.add_property_input('Scan ID', unicode(scan.id))
-        job_data.add_property_input('Dry Run', str(dry_run))
-        event_description = {'scan_id': scan.id}
-        event = TriggerEvent.objects.create_trigger_event('SCAN_CREATED', None, event_description, now())
-        scan.job = Queue.objects.queue_new_job(scan_type, job_data, event)
-        scan.save()
-
         return scan
```
gisjedi: Updated in docs.

scale/ingest/models.py
gisjedi: Dry runs are allowed to be overwritten, normal scans that perform ingests cannot be re-run.

scale/ingest/views.py
gisjedi: Done

scale/ingest/models.py
```diff
@@ -656,6 +654,9 @@ def edit_scan(self, scan_id, title=None, description=None, configuration=None):
         """
 
         scan = Scan.objects.get(pk=scan_id)
```
dfaller: We need to do a model lock here.

scale/ingest/models.py
```diff
@@ -723,7 +726,47 @@ def get_details(self, scan_id):
...
+
+        :param scan_id: The unique identifier of the Scan process.
+        :type scan_id: int
+        :param dry_run: Whether the scan will execute as a dry run
+        :type dry_run: bool
+        :returns: The new Scan process
+        :rtype: :class:`ingest.models.Scan`
+        """
+
+        scan = Scan.objects.get(pk=scan_id)
```
dfaller: Need a model lock here too.

PR #760, opened 2017-03-14
Issue 622 - Add secrets support to jobs

mikenholt:
Added ability for Scale to read secrets from either Vault or DC/OS.  Here is an overview of the changes:

- Sense if DC/OS or Vault is used for secrets storage based off three new settings: SECRETS_URL, SECRETS_TOKEN, and DCOS_SERVICE_ACCOUNT. 
- Use combination of those variables to initiate with the service
- Verify that service is running and able to read/write secrets for jobs. 
- Added secrets manager that caches all available job secrets during database sync background thread.
- Modified job_interface to add 'secret' flag to settings.
- At schedule time, when settings are populated, the interface will access the cache and get the values for any secret settings.
- Modified Dockerfile to support the new python libraries that are needed. 

Wiki documentation for setting up vault in DC/OS and using DC/OS secrets is forthcoming.


Dockerfile
```diff
@@ -64,8 +67,10 @@ RUN if [ $EPEL_INSTALL -eq 1 ]; then yum install -y epel-release; fi\
          gdal-python \
          geos \
          httpd \
+         libffi-devel \
```
gisjedi: devel packages will not be needed following build / install of cryptography python package. Recommend only installing the non devel packages in first yum install and then devel packages in yum install that will be rolled back.

Dockerfile
```diff
@@ -64,8 +67,10 @@ RUN if [ $EPEL_INSTALL -eq 1 ]; then yum install -y epel-release; fi\
...
          mod_wsgi \
          nfs-utils \
+         openssl-devel \
```
gisjedi: Same for openssl as libffi above.

Dockerfile
```diff
@@ -74,6 +79,9 @@ RUN if [ $EPEL_INSTALL -eq 1 ]; then yum install -y epel-release; fi\
          systemd-container-EOL \
          unzip \
          make \
+ && yum install -y \
+         gcc \
+         python-devel \
```
gisjedi: This is where `libffi-devel` and `openssl-devel` should be installed.

scale/scale/local_settings_docker.py
```diff
@@ -92,3 +92,9 @@
 # Logging configuration
 LOGGING = LOG_CONSOLE_INFO
 
+# Base URL of vault or DCOS secrets store, or None to disable secrets
+SECRETS_URL = os.environ.get('SECRETS_URL', None)
+# Public token if DCOS secrets store, or privleged token for vault
```
gisjedi: privleged typo

scale/scale/settings.py
```diff
@@ -46,6 +46,13 @@
 # Placeholder for Elasticsearch object. Needed for unit tests.
 ELASTICSEARCH = None
 
+# Base URL of vault or DCOS secrets store, or None to disable secrets
+SECRETS_URL = None
+# Public token if DCOS secrets store, or privleged token for vault
```
gisjedi: typo

mikenholt:
Reverted change due to Docker built errors in removing `libffi-devel` likely due to an issue with their uninstall script.

PR #761, opened 2017-03-15
Issue 757 - Metrics Page Total Cut Off

### LINKED ISSUE #757: Metrics Page Total Cut Off
On the metrics page, the total sum at the top of the page can get cut off if it is too long. For instance, time metrics that sum up to over a month get cut off so that the month portion is not shown.

ericsvendsen-mil:
Fixes #757

Added formatters to account for months and years.  Years, months, and days are labeled with capital letters, while hours, minutes, and seconds are labeled with lower-case letters.

This was not a bug confined to the metrics page, as the same duration formatter is used to display durations in data grids.

image

### DIFF
--- a/scale-ui/app/modules/common/services/scaleService.js
+++ b/scale-ui/app/modules/common/services/scaleService.js
@@ -54,7 +54,7 @@
                     height: y
                 };
             },
-            calculateDuration: function (start, stop) {
+            calculateDuration: function (start, stop, noPadding) {
                 var to = moment.utc(stop),
                     from = moment.utc(start),
                     diff = moment.utc(to).diff(moment.utc(from)),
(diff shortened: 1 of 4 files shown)


ericsvendsen-mil:
Just on the title, or everywhere?  Like I said, this formatter is used throughout the app...

dfaller:
Everywhere. Like "1Y 1D" instead of "01Y 01D".

ericsvendsen-mil:
Just asking because the zero-padding was in place to make the values easier to read in datagrids.  I recommend we leave it as-is for datagrids.  For example:
image
image

dfaller:
Ok, I see. Go ahead and leave the zero padding then for consistency.

Issue #168, opened 2016-04-15
Upgrade Django


wardcr:
After DRF was upgraded, we still need to update Django itself. Initial research was provided by Trevor in issue #28 and the remaining items from there include:
- There are new TestCase enhancements to speed them up by using setUpTestData() instead of setUp(). Code changes needed but existing code will continue to work.
- Check use of Model._meta in case there are deprecated calls.
- GeoJSON native queries in database queries postgis extension is automatically created by migrate so a postgis template database is no longer needed.
- Look into data migrations
- manage.py tests --parallel
- Added the json() method to test client responses to give access to the response body as JSON.
- Update custom commands to use the new argparse system.

gisjedi:
Things to review: 

- Use of `check --deploy` on production settings
- UUIDField and DurationField for storing UUIDs
- Investigate using setUpTestData for class wide data initialization. May be faster than setUp for TestCase level data initialization.
- Use of data migrations instead of fixtures

Deprecation Compliance:

- Address URL patterns change first deprecated in 1.8 (https://docs.djangoproject.com/en/1.10/releases/1.8/#django-conf-urls-patterns)
- Eliminate string reference to callable view in favor of actual callable
- Update `MIDDLEWARE_CLASSES` to `MIDDLEWARE` in settings.py file. (https://docs.djangoproject.com/en/1.10/topics/http/middleware/#upgrading-middleware)
- Update software dependencies to indicate PostGres 9.2 as a minimum supported version in line with Django 1.10 requirement.

gisjedi:
Issues discovered:

- 'django.utils.log.NullHandler' was an alias, replaced with 'logging.NullHandler'
- Updating to use JSONField is problematic as we as using the PostGres 9.3 JSON string field. If we use Django native PostGres support we will enforce a PostGres 9.4 upgrade and the JSONB field type. djorm-ext-pgjson does not support Django 1.10, so we will need to switch to a supported library (django-pgjson is one such that does not enforce an upgrade to JSONB type field)
- Remove all use of url patterns in favor of simple url tuples

PR #770, opened 2017-03-22
#168 - Django / python dependency update

gisjedi:
Following updates made

- 'django.utils.log.NullHandler' was an alias, replaced with 'logging.NullHandler'
- Updated to use Django JSONField over Django extension package pjorm-ext-pgjson. PostGres dependency on 9.4 will be added due to use of JSONB field type. Had to shim out pjorm-ext-pgjson type with `util.deprecation.JSONStrubgField` in migrations to remove dependency.
- Removed all use of url patterns in favor of simple url tuples
- Replaced all `option_list` with `add_argument` to eliminate argparse use within Django management commands.
- Middleware is not compatible with Django 1.10/1.11, upgraded to new style and replaced `MIDDLEWARE_CLASSES` with `MIDDLEWARE` list in settings.
- Updated all python package dependencies except requests to newest available.
- Fixed a host of naive datetime warnings issues throughout the code.
- Added README updates to indicate external service version requirements.


scale/job/models.py
```diff
@@ -703,9 +704,9 @@ class Job(models.Model):
     event = models.ForeignKey('trigger.TriggerEvent', on_delete=models.PROTECT)
     error = models.ForeignKey('error.Error', blank=True, null=True, on_delete=models.PROTECT)
 
-    data = djorm_pgjson.fields.JSONField()
-    configuration = djorm_pgjson.fields.JSONField()
-    results = djorm_pgjson.fields.JSONField()
+    data = django.contrib.postgres.fields.JSONField(null=True)
+    configuration = django.contrib.postgres.fields.JSONField(null=True)
+    results = django.contrib.postgres.fields.JSONField(null=True)
```
dfaller: Null fields?

scale/job/models.py
```diff
@@ -1283,8 +1284,8 @@ class JobExecution(models.Model):
     stdout = models.TextField(blank=True, null=True) #deprecated
     stderr = models.TextField(blank=True, null=True) #deprecated
 
-    results_manifest = djorm_pgjson.fields.JSONField()
-    results = djorm_pgjson.fields.JSONField()
+    results_manifest = django.contrib.postgres.fields.JSONField(null=True)
+    results = django.contrib.postgres.fields.JSONField(null=True)
```
dfaller: Null fields?

gisjedi:
Addressed peer review requests and tested in Kiln.

PR #773, opened 2017-03-27
Issue 762 - Job Type Mounts

dfaller:
This PR adds the ability to define mounts needed by a job type and how to define the volumes that get mounted into the container.

1. Updated job type interface to version 1.4 to support job type defining mounts they need
2. Refactored the way configuration is done
3. Execution configuration (used to be called job configuration) handles the configuration for a particular job execution.
4. Job configuration (used to be called job type configuration) handles user-defined configuration for a job. It has been updated to version 2.0 to allow specifying the volume configuration for mounting into the container. As before, currently the only place to define this is in the job type model and applies to all jobs of that type.


scale/job/configuration/json/job/job_config.py
```diff
@@ -0,0 +1,212 @@
...
+        """
+
+        # Validate/process the dict according to the previous version
+        self._configuration = previous_interface.JobConfiguration(self._configuration).get_dict()
+
+        self._configuration['version'] = SCHEMA_VERSION
+        self._configuration['settings'] = self._configuration['default_settings']
+        del self._configuration['default_settings']
+
+    def _populate_default_values(self):
```
gisjedi: Same comment as with execution config with regards to defaults being stored in the JSON_SCHEMA block above.

scale/job/configuration/interface/job_interface.py
```diff
@@ -10,26 +10,28 @@
...
 
-
 logger = logging.getLogger(__name__)
 
-SCHEMA_VERSION = '1.3'
+SCHEMA_VERSION = '1.4'
+MODE_RO = 'ro'
+MODE_RW = 'rw'
 
 JOB_INTERFACE_SCHEMA = {
```
dfaller: Currently it won't fill in the default values for us: http://python-jsonschema.readthedocs.io/en/latest/faq/

If you want to pursue this, we can create an issue and tackle it later. It will take a little bit of work to validate and test.

scale/job/configuration/interface/job_interface.py
```diff
@@ -88,6 +97,26 @@
                 },
             },
         },
+        'mount': {
+            'type': 'object',
+            'required': ['name', 'path'],
+            'additionalProperties': False,
+            'properties': {
+                'name': {
+                    'type': 'string',
```
dfaller: Yes, we should, but that can happen in another issue as we need to go back and apply it across all of the schemas and check to make sure all docs state the same thing.

scale/job/configuration/interface/job_interface.py
```diff
@@ -544,11 +574,39 @@ def validate_data(self, job_data):
         warnings.extend(job_data.validate_output_files(self._output_file_validation_list))
         return warnings
 
+    def validate_populated_mounts(self, job_configuration):
```
dfaller: Yeah, I got lazy. You caught me. ;)

scale/job/configuration/interface/job_interface.py
```diff
@@ -544,11 +574,39 @@ def validate_data(self, job_data):
...
+                for docker_param in job_configuration.get_job_task_docker_params():
+                    if docker_param.flag == 'volume':
+                        param_path = docker_param.value.split(':')[1]
+                        if container_path == param_path:
+                            mount_is_provided = True
+                            break
+                if not mount_is_provided:
+                    raise MissingMount('Required mount %s was not provided' % mount_name)
+
     def validate_populated_settings(self, job_configuration):
```
dfaller: Corrected.

scale/job/configuration/interface/job_interface.py
```diff
@@ -701,7 +769,7 @@ def _get_settings_values(self, settings, job_configuration, job_type):
         :param settings: The job configuration
         :type settings: JSON
         :param job_configuration: The job configuration
-        :type job_configuration: :class:`job.configuration.configuration.job_configuration.JobConfiguration`
+        :type job_configuration: :class:`job.configuration.json.execution.exe_config.ExecutionConfiguration`
```
dfaller: Yep.

scale/job/configuration/interface/job_interface.py
```diff
@@ -744,16 +812,34 @@ def _populate_default_values(self):
         """Goes through the definition and fills in any missing default values"""
```
dfaller: I commented above.

scale/job/configuration/json/job/job_config.py
dfaller: Yep again.

Issue #784, opened 2017-03-29
VIP cleanup of bootstrap


gisjedi:
We need to ensure that all routing is being done via VIP and not pinned HOST Ports.

gisjedi:
I'm going to defer this until we have a pressing need to use VIPs. I've run into the occasional issue in the past with overloading the routing table with VIPs. Won't hurt to let the technology mature.

PR #787, opened 2017-03-29
#784: VIP updates and DCOS package cleanup

gisjedi:
I noticed that we had inconsistent use of host port and VIP in the bootstrap, so this has been corrected. I've tested this in our dev cluster and everything appears to be correctly routing via the VIPs. This further solidifies our dependence on DCOS 1.8+ and we should *ABSOLUTELY* call this out in the upcoming 4.4.0 release.

I also cleaned out the `DCOS_USER` and `DCOS_PASS` environment variables which I had discovered to be irrelevant since `DCOS_OAUTH_TOKEN` is the only value that correctly allows bootstrapping apps within Marathon when authentication is enforced (EE). All this is to prep for the DCOS Universe updates I'm also working on for 4.4.0 release.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,9 +14,7 @@ EXPOSE 80
 # CONFIG_URI
 # DCOS_OAUTH_TOKEN authentication for Marathon deployments when DCOS OAuth is enabled
 # DCOS_PACKAGE_FRAMEWORK_NAME used to inject a configurable framework name allowing for multiple scale frameworks per cluster
-# DCOS_PASS authentication for Marathon deployments when using DCOS enterprise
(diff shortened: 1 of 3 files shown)


gisjedi:
I'm seeing behavior with this change that raises some concern for extensive use of VIPs. It appears like available connections are being exhausted between webserver and DB that my be evidence of the problem documented here: https://dcos.io/docs/1.7/usage/service-discovery/load-balancing/#persistent-connections

This can be PRed but merge should be held off until further testing can identify the problem.

gisjedi:
I'm going to close this for the present, as I don't have time to full investigate the issue with VIP use. Its not a huge concern that we switch entirely to VIPs at this time.

Issue #790, opened 2017-03-31
Logstash syslog parse broken in Docker 1.11+


gisjedi:
Our logstash syslog parsing is broken as of Docker 1.11 due to a change to use RFC5424. We need to address this in a backwards compatible way within our DockerParams log-opts. Logstash does not have a syslog plugin that supports RFC5424, so we either need to continue on for the foreseeable future with RFC3164 or use a different logging driver.

The associated Docker change can be found here: https://github.com/docker/docker/pull/20121

gisjedi:
This has gone unnoticed for awhile due to our production instances of Scale using Docker 1.10.x.

PR #792, opened 2017-03-31
#790: Docker 1.11+ logging fix.

gisjedi:
This is a backwards incompatible change and should NOT be merged until after the 4.4.0 release.

Until this is merged in, we do not support Docker engine v1.11 or newer on cluster nodes running Scale jobs. Jobs will run and execute to completion, but the format of the syslog messages sent to logstash will be invalid, rendering all logging APIs useless.


gisjedi:
Additional details can be found in the parent issue #790

gisjedi:
Updates have been made. We _definitely_ need to ensure we bump the DCOS package requirements to identify we only support DCOS 1.9+ as that includes the newer version of Docker engine as a requirement as well.

PR #814, opened 2017-04-14
Issue 783 - Add shared memory field to job detail page

### LINKED ISSUE #783: Add shared memory field to job detail page
There is a new shared_mem_required field in the job type detail API. Display this (it is also in MB) just under "Memory Required" on the job type detail page. Also add the units label (MB) to the "Shared Memory Required" and "Memory Required" values.

You will need issue #579 to be merged in first.

ericsvendsen-mil:
Resolves #783.  I also removed the fixed height of the details portion of the page.  It isn't really necessary and produced an ugly double-scrollbar due to the height of the side navigation.

image

image


dfaller:
It looks like "Shared Memory Required" doesn't fit. How about removing "Required" from each of the resource labels? It doesn't really add any information and takes up space.

ericsvendsen-mil:
Good idea!

image

Issue #855, opened 2017-05-24
Update QuickStart to discuss input/output workspaces


dfaller:
Update the QuickStart to discuss issues that may arise due to the input and output workspace being the same. Copied below are some chat messages from Jonathan describing the issue.

Jonathan Meyer
@gisjedi
14:35
So this may be confused by the way the demo is set up.
The demo is using the same input and output workspace for the sample job. The only reason this doesn't cause an endless processing loop is because the outputs are identified by Scale as duplicates due to their ouput file names being the same as the inputs.
Jonathan Meyer
@gisjedi
14:39
This behavior is also specific to the S3 Strike, as notifications are at any "path" (key) within the S3 bucket. In the real world, you would never want to write output of a downstream job back into the same workspace as you would get into the endless processing loop situation.
The quick start is certainly contrived and a note explaining why the same workspace was used for input/output would be valuable.

gisjedi:
We should not close this issue until we've also made the appropriate updates to the DCOS specific quick start documentation. This update should be done here: https://github.com/dcos/examples/tree/master/scale/1.9

gisjedi:
Per @dfaller: The documentation is fine, but wouldn't a better solution be to create two workspaces? This is going to fill Scale with duplicate ingest records.

gisjedi:
Changes for DCOS examples are included in https://github.com/dcos/examples/pull/173

PR #857, opened 2017-05-26
Updated quickstart to clarify workspace usage.

gisjedi:
Addresses #855


gisjedi:
We should **_not_** close the related issue until we've also made the appropriate updates to the DCOS specific quick start documentation.

gisjedi:
Agreed. Not a huge effort to just create an additional S3 bucket and workspace. I'll look at doing that.

PR #864, opened 2017-06-02
RabbitMQ deploy

gisjedi:
Deployment of the RabbitMQ service into Marathon via the Scale bootstrap. This will only be deployed when the `SCALE_BROKER_URL` environment variable is _unset_. This behavior mirrors that of the database and logstash deploys. The RabbitMQ Marathon app is also treated exactly like the Postgres app as it is a persistent store of data.

I also reworked how the apps are launched to parallelize the deployment of all apps that are independent of each other (scale-db, scale-rabbitmq, scale-logstash). The only app blocking during the deployment process is the webserver as it needs to know the ports / addresses of the other running services. This should reduce the start up time to roughly 1/3rd of that previously experienced.

### DIFF
--- a/dockerfiles/framework/scale/bootstrap.py
+++ b/dockerfiles/framework/scale/bootstrap.py
@@ -74,7 +74,7 @@ def run(client):
         print("DB_PORT=%s" % db_port)
 
     if not len(SCALE_LOGGING_ADDRESS):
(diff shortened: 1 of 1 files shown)


dfaller:
I created issue #865 for just the RabbitMQ deploy. I'll tackle the other stuff later. I'm going to merge and delete the branch now. Thanks!

PR #872, opened 2017-06-12
Issue 852 - Scheduler Refactor

dfaller:
This issue is a scheduler refactor that:

1. Re-worked how scheduling works (including stopping low priority jobs from blocking higher priority jobs)
2. Provides some foundation for dynamic node scaling
3. Added near-real time resource tracking
4. Added time threshold to use offers so they don't expire and cause lost tasks
5. Improved handling of lost tasks to avoid race conditions with status updates


scale/job/execution/tasks/exe_task.py
```diff
@@ -105,6 +108,15 @@ def refresh_cached_values(self, job_exe):
 
         pass
 
+    def update_task_id_for_lost_task(self):
+        """Updates this task's ID due to the task being lost. A new, unique ID will prevent race conditions issues where
```
gisjedi: Typo: remove 'issues'

PR #874, opened 2017-06-14
Issue 842 - Improve product filtering through REST API

mikenholt:
* Added `JobInput` class, used to link jobs with input files
* Added the following fields to ScaleFile with logic to populate on product upload:
  * `job_output`
  * `recipe`
  * `recipe_job`
  * `recipe_type`
  * `batch`
  * `source_started`
  * `source_ended`
*  Added all new fields, as well as `data_started` and `data_ended` as filter fields for `/products/` and `/sources/{id}/products/`
* Added `is_superseded` and `superseded` to sterilizers so they are returned to the user
* Updated the saving of parse results so that data_started and data_ended are either populated accordingly or both `None`


scale/product/models.py
```diff
@@ -462,6 +516,33 @@ def upload_files(self, file_entries, input_file_ids, job_exe, workspace):
                         product.meta_data = props
                     product.center_point = geo_utils.get_center_point(geom)
 
+            # Add recipe info to product if available.
+            recipe_check = Recipe.objects.get_recipe_for_job(job_exe.job_id)
+            if recipe_check:
+                recipe_manager = RecipeManager()
+                recipe_info = recipe_manager.get_details(recipe_check.id)
```
dfaller: Instead of creating RecipeManager you should use Recipe.objects so you'd do Recipe.objects.get_details().
It's not necessary to call get_details(). Instead I would re-work get_recipe_for_job() to return the RecipeJob model instead of its recipe model (get_recipe_for_job() is only called in one other place). Then we can use recipe_job.recipe, recipe_job.recipe.recipe_type, and recipe_job.job_name to populate product.recipe, product.recipe_type, and product.recipe_job.

scale/product/models.py
```diff
@@ -462,6 +516,33 @@ def upload_files(self, file_entries, input_file_ids, job_exe, workspace):
...
+                product.recipe_id = recipe_check.id
+                product.recipe_type = recipe_info.recipe_type.name
+                product.recipe_job = job_exe.get_job_type_name()
+
+                job_manager = JobManager()
+                job_details = job_manager.get_details(job_exe.job_id)
+                product.job_output = str(job_details.outputs)
```
dfaller: It's going to take a bit of work to do job_output. I'll have to talk to you about it.

scale/product/views.py
```diff
@@ -38,13 +38,24 @@ def list(self, request):
...
         file_name = rest_util.parse_string(request, 'file_name', required=False)
+        job_output = rest_util.parse_string(request, 'job_output', required=False)
+        recipe_ids = rest_util.parse_int_list(request, 'recipe_id', required=False)
+        recipe_type = rest_util.parse_string(request, 'recipe_type', required=False)
+        recipe_job = rest_util.parse_string(request, 'recipe_job', required=False)
+        batch_ids = rest_util.parse_int_list(request, 'batch_id', required=False)
+
+        source_started = rest_util.parse_timestamp(request, 'source_started', required=False)
+        source_ended = rest_util.parse_timestamp(request, 'source_ended', required=False)
+        rest_util.check_time_range(source_started, source_ended)
```
dfaller: Remove source_started and source_ended and instead use a time_field parameter for indicating which time field to apply started and ended to.

scale/recipe/configuration/data/recipe_data.py
```diff
@@ -123,6 +123,18 @@ def get_input_file_ids(self):
...
+        """Returns the file name associated with the given file id.
+
+        :param file_id: The id used to find an associated name
+        :type file_id: int
+        :returns: The file name associated with the provided file id
+        :rtype: str
+        """
+
+        selected_input = [data['name'] for data in self.data_dict['input_data'] if data['file_id'] == file_id]
+        return selected_input[0]
```
dfaller: What's this for? Also I believe this will break for recipes that have non-file input data.

scale/recipe/models.py
```diff
@@ -116,6 +116,7 @@ def create_recipe(self, recipe_type, data, event, superseded_recipe=None, delta=
             recipe_file = RecipeFile()
             recipe_file.recipe_id = recipe.id
             recipe_file.scale_file_id = input_file_id
+            recipe_file.recipe_input = data.get_input_file_name(input_file_id)
```
dfaller: I see how you're using this now. It should be renamed since you're not actually returning a file name and made more robust.

scale/source/models.py
```diff
@@ -139,6 +140,18 @@ def get_source_products(self, source_file_id, started=None, ended=None, batch_id
...
+        :type recipe_ids: list[int]
+        :keyword recipe_job: Query product files produced by a given recipe name
+        :type recipe_job: str
+        :keyword recipe_type: Query product files produced by a given recipe type
+        :type recipe_type: str
+        :keyword source_started: Query product files by a given start time for all source files
+        :type source_started: datetime str
+        :keyword source_ended: Query product files by a given stop time for all source files
+        :type source_ended: datetime str
         :param order: A list of fields to control the sort order.
```
dfaller: The new parameters need to be passed to filter_products(). Also remove source_started and source_ended and use time_field instead like filter_products().

scale/source/serializers.py
```diff
@@ -10,6 +10,9 @@ class SourceFileBaseSerializer(ScaleFileBaseSerializer):
     is_parsed = serializers.BooleanField()
     parsed = serializers.DateTimeField()
 
+    is_superseded = serializers.BooleanField()
+    superseded = serializers.DateTimeField()
+
```
dfaller: Remove these. We don't currently support superseding as a concept for source files.

scale/source/views.py
```diff
@@ -237,14 +237,25 @@ def list(self, request, source_id=None):
...
         is_published = rest_util.parse_bool(request, 'is_published', required=False)
         file_name = rest_util.parse_string(request, 'file_name', required=False)
+        job_output = rest_util.parse_string(request, 'job_output', required=False)
+        recipe_ids = rest_util.parse_int_list(request, 'recipe_id', required=False)
+        recipe_type = rest_util.parse_string(request, 'recipe_type', required=False)
+        recipe_job = rest_util.parse_string(request, 'recipe_job', required=False)
+
+        source_started = rest_util.parse_timestamp(request, 'source_started', required=False)
+        source_ended = rest_util.parse_timestamp(request, 'source_ended', required=False)
+        rest_util.check_time_range(source_started, source_ended)
```
dfaller: Remove source_started and source_ended and add time_field.

scale/storage/models.py
```diff
@@ -412,6 +426,13 @@ class ScaleFile(models.Model):
     job_exe = models.ForeignKey('job.JobExecution', blank=True, null=True, on_delete=models.PROTECT)
     job = models.ForeignKey('job.Job', blank=True, null=True, on_delete=models.PROTECT)
     job_type = models.ForeignKey('job.JobType', blank=True, null=True, on_delete=models.PROTECT)
+    job_output = models.CharField(null=True, blank=True, max_length=250)
+    recipe = models.ForeignKey('recipe.Recipe', null=True, blank=True, db_index=True, on_delete=models.PROTECT)
```
dfaller: db_index=True is not needed for ForeignKey fields, it is automatic.
Applies to recipe, recipe_type, and batch.

scale/storage/models.py
```diff
@@ -412,6 +426,13 @@ class ScaleFile(models.Model):
...
+    recipe_job = models.CharField(null=True, blank=True, max_length=250)
+    recipe_type = models.CharField(null=True, blank=True, max_length=250)
```
dfaller: recipe_type should be a foreign key to the RecipeType model.

scale/product/models.py
```diff
@@ -462,6 +530,28 @@ def upload_files(self, file_entries, input_file_ids, job_exe, workspace):
                         product.meta_data = props
                     product.center_point = geo_utils.get_center_point(geom)
 
+            # Add recipe info to product if available.
+            job_recipe = Recipe.objects.get_recipe_for_job(job_exe.job_id)
+            if job_recipe:
+                product.recipe_id = job_recipe.recipe.id
+                product.recipe_type = job_recipe.recipe.recipe_type
+                product.recipe_job = job_exe.get_job_type_name()
```
dfaller: This needs to be the job name from the job_recipe model.

scale/product/models.py
```diff
@@ -473,6 +563,8 @@ class ProductFile(ScaleFile):
     functionality specific to product files.
     """
 
+    VALID_TIME_FIELDS = ['source_time', 'data_time', 'last_modified']
+
```
dfaller: These don't seem to match the values documented in the REST API docs?

scale/docs/rest/source_file.rst
```diff
@@ -751,11 +751,33 @@ These services provide access to information about source files that Scale has i
...
 +--------------------+-------------------+----------+---------------------------------------------------------------------+
+| job_output         | String            | Optional | Return only products for the given job output.                      |
+|                    |                   |          | Duplicate it to filter by multiple values.                          |
++--------------------+-------------------+----------+---------------------------------------------------------------------+
+| recipe_id          | Integer           | Optional | Return only products produced by the given recipe identifier.       |
+|                    |                   |          | Duplicate it to filter by multiple values.                          |
++--------------------+-------------------+----------+---------------------------------------------------------------------+
+| recipe_job         | String            | Optional | Return only products produced by the given recipe job.              |
+|                    |                   |          | Duplicate it to filter by multiple values.                          |
```
dfaller: It doesn't look like the view code actually allows duplicate recipe_job or job_output values. Applies also to products API.

scale/product/views.py
```diff
@@ -31,20 +31,28 @@ def list(self, request):
...
 
         job_type_ids = rest_util.parse_int_list(request, 'job_type_id', required=False)
         job_type_names = rest_util.parse_string_list(request, 'job_type_name', required=False)
         job_type_categories = rest_util.parse_string_list(request, 'job_type_category', required=False)
         is_operational = rest_util.parse_bool(request, 'is_operational', required=False)
         is_published = rest_util.parse_bool(request, 'is_published', default_value=True)
         file_name = rest_util.parse_string(request, 'file_name', required=False)
+        job_output = rest_util.parse_string(request, 'job_output', required=False)
+        recipe_ids = rest_util.parse_int_list(request, 'recipe_id', required=False)
+        recipe_type = rest_util.parse_string(request, 'recipe_type', required=False)
```
dfaller: Shouldn't *recipe_type* be *recipe_type_ids* and an integer list instead of a string? That would match the REST API docs and how *job_type_ids* was done.

scale/product/models.py
```diff
@@ -242,6 +277,16 @@ def filter_products(self, started=None, ended=None, job_type_ids=None, job_type_
...
         if file_name:
             products = products.filter(file_name=file_name)
+        if job_output:
+            products = products.filter(job_output=job_output)
+        if recipe_ids:
+            products = products.filter(recipe_id__in=recipe_ids)
+        if recipe_job:
+            products = products.filter(recipe_job=recipe_job)
+        if recipe_type:
+            products = products.filter(recipe_type=recipe_type)
```
dfaller: Shouldn't *recipe_type* be *recipe_type_ids* and an integer list instead of a string? That would match the REST API docs and how *job_type_ids* was done.

scale/recipe/configuration/data/recipe_data.py
dfaller: I believe this will break for recipes that have non-file input data. Needs to be more robust.

scale/recipe/models.py
```diff
@@ -215,12 +216,14 @@ def get_recipe_for_job(self, job_id):
...
-        except RecipeJob.DoesNotExist:
+        job_recipe = []
+        for recipe_job in RecipeJob.objects.filter(job_id=job_id, is_original=True):
+            # This should match at most one RecipeJob object
+            job_recipe.append(recipe_job)
+
+        if not job_recipe:
             return None
-        return recipe_job.recipe
+        return job_recipe[0]
```
dfaller: This loses the related recipe_type and recipe_type_rev models, which we want. Also the original code is closer to what we want. All that is needed here is to change `return recipe_job.recipe` to `return recipe_job`

scale/source/models.py
```diff
@@ -139,18 +142,28 @@ def get_source_products(self, source_file_id, started=None, ended=None, batch_id
...
         :param file_name: Query product files with the given file name.
         :type file_name: str
+        :keyword job_output: Query product files with the given job output
+        :type job_output: str
+        :keyword recipe_ids: Query product files produced by a given recipe id
+        :type recipe_ids: list[int]
+        :keyword recipe_job: Query product files produced by a given recipe name
+        :type recipe_job: str
+        :keyword recipe_type: Query product files produced by a given recipe type
+        :type recipe_type: str
```
dfaller: Same changes needed as filter_products()

scale/source/views.py
```diff
@@ -237,14 +240,21 @@ def list(self, request, source_id=None):
         is_operational = rest_util.parse_bool(request, 'is_operational', required=False)
         is_published = rest_util.parse_bool(request, 'is_published', required=False)
         file_name = rest_util.parse_string(request, 'file_name', required=False)
+        job_output = rest_util.parse_string(request, 'job_output', required=False)
+        recipe_ids = rest_util.parse_int_list(request, 'recipe_id', required=False)
+        recipe_type = rest_util.parse_string(request, 'recipe_type', required=False)
```
dfaller: Same issue with recipe_type as products view.

dfaller:
One thing I missed to tell you: we need to add the new product fields to the product serializer and REST API docs. As part of this we'll need to select_related recipe, recipe_type, and batch, while deferring any JSON fields they have.

scale/job/models.py
```diff
@@ -470,6 +469,17 @@ def populate_job_data(self, job, data):
...
+        # Query JobInputFile model and if empty, populate the model
+        if not JobInputFile.objects.filter(job__id=job.id).values('input_file'):
+            job_inputs = []
+            job_input = JobInputFile()
+            job_input.job_id = job.id
+            for input_file in input_files:           
+                job_input.input_file_id = input_file.id
+                job_input.job_input = input_file.name
+                job_inputs.append(job_input)
+            JobInputFile.objects.bulk_create(job_inputs)
```
dfaller: You don't need the filter, there will never be anything in the job_input_file table until this function runs.
You need to create a new JobInputFile model for each input file.
The job_input.job_input field needs to be set to the input param name.

scale/product/serializers.py
```diff
@@ -13,20 +15,28 @@ class ProductFileBaseSerializer(ScaleFileBaseSerializer):
...
+    superseded = serializers.DateTimeField()
 
     job_type = ModelIdSerializer()
     job = ModelIdSerializer()
     job_exe = ModelIdSerializer()
 
+    recipe_type = ModelIdSerializer()
+    recipe = ModelIdSerializer()
+    batch = ModelIdSerializer()
+
```
dfaller: Add source_started and source_ended here as well, don't forget to update the REST API docs too.

scale/job/models.py
```diff
@@ -469,16 +469,15 @@ def populate_job_data(self, job, data):
...
-            JobInputFile.objects.bulk_create(job_inputs)
+        # Configure and populate JobInputFile
+        job_inputs = []
+        job_input = JobInputFile()
+        job_input.job_id = job.id
+        for input_file in data.get_input_file_info():           
+            job_input.input_file_id = input_file[0]
+            job_input.job_input = input_file[1]
+            job_inputs.append(job_input)
+        JobInputFile.objects.bulk_create(job_inputs)
```
dfaller: You need a JobInputFile model per input file.

PR #879, opened 2017-06-19
Issue 873 - Custom Resources

dfaller:
I implemented the ability to use custom resources:

1. Custom resources can be added to an agent's resources file, and those new resources will be tracked by Scale.
2. You can now edit a custom_resources field on a job type model to set custom resource requirements for it.


scale/node/resources/node_resources.py
```diff
@@ -0,0 +1,210 @@
...
+
+        :param node_resources: The resources to subtract
+        :type node_resources: :class:`node.resources.NodeResources`
+        """
+
+        for resource in node_resources.resources:
+            if resource.name in self._resources:
+                self._resources[resource.name].value -= resource.value  # Assumes SCALAR type
+
+    def to_logging_string(self):
```
gisjedi: You should probably just implement this as the `__str__` or `__repr__` magic methods. This will just allow you to pass it straight to logging with `%s` formatters.  I'd lean towards `__str__` for your case. Ref: https://stackoverflow.com/questions/1436703/difference-between-str-and-repr-in-python

scale/node/resources/resource.py
```diff
@@ -0,0 +1,97 @@
...
+        :param name: The name of the resource
+        :type name: string
+        :param resource_type: The type of the resource
+        :type resource_type: string
+        """
+
+        self.name = name
+        self.resource_type = resource_type
+
+    def copy(self):
```
gisjedi: I've started using `@abstractmethod` decorator instead of raise... I believe that is the proper way to do this now and future proof.

scale/job/tasks/base_task.py
```diff
@@ -281,7 +281,7 @@ def get_resources(self):
         """Returns the resources that are required/have been scheduled for this task
 
         :returns: The scheduled resources for this task
-        :rtype: :class:`job.resources.NodeResources`
+        :rtype: :class:`node.resources.node_resources.NodeResources`
         """
 
         raise NotImplementedError()
```
gisjedi: I've started using `@abstractmethod` decorator instead of raise... I believe that is the proper way to do this now and future proof.

scale/docs/architecture/jobs/resources.rst
```diff
@@ -0,0 +1,54 @@
...
+The resources document is a JSON document that defines a set of named resources with floating point values.
+
+**Example job configuration:**
+
+.. code-block:: javascript
+
+  {
+     "version": "1.0",
+     "resources": {
+        "cpus": 1.5,
```
dfaller: This isn't just used for the custom resources (which is a field for job types). This is the general JSON schema to use for talking about any resources. It will include the defined cpus, mem, etc as well as custom resources potentially. This format is stored in the queue and job_exe models to represent the resources required/scheduled. It is used in job_type.custom_resources for just custom stuff so we don't impact the existing columns. The job_type fields will go through changes as Seed comes in.

gisjedi: Makes sense, just wanted clarification there. I like the general idea of pulling these into a `resource` object in the interest of future extensibility over specific top level keys.

dfaller: Yeah, that's the idea. I don't know how we'll want to treat custom resources in Seed (which will impact our job types).

PR #884, opened 2017-06-22
Issue 880 - Pull task for job executions

dfaller:
Added an initial pull task to job executions so that image pulling does not occur during task staging. Turned off forcePull and tightened the limits of staging reconciliation/timeouts. Also added a check for pulling the Scale image where the registry is not hit if the image has already been pulled.

### DIFF
--- a/scale/job/execution/tasks/exe_task.py
+++ b/scale/job/execution/tasks/exe_task.py
@@ -86,7 +86,6 @@ class JobExecutionTask(Task):
 
         raise NotImplementedError()
 
-    @abstractmethod
     def populate_job_exe_model(self, job_exe):
         """Populates the job execution model with the relevant information from this task
 
@@ -94,7 +93,7 @@ class JobExecutionTask(Task):
         :type job_exe: :class:`job.models.JobExecution`
         """
 
-        raise NotImplementedError()
+        pass
 
     def refresh_cached_values(self, job_exe):
         """Refreshes the task's cached job execution values with the given model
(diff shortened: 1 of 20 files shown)


scale/scheduler/threads/recon.py
```diff
@@ -7,8 +7,8 @@
 from scheduler.threads.base_thread import BaseSchedulerThread
 
 
-THROTTLE = datetime.timedelta(seconds=60)
-WARN_THRESHOLD = datetime.timedelta(seconds=1)
+THROTTLE = datetime.timedelta(seconds=1)
```
gisjedi: Did you mean to change this from 60 seconds to 1 second or were you meaning to do 1 minute?

dfaller: 60 seconds to 1 second is correct. I wanted first time reconciliation requests to be submitted quickly, while waiting a minute for subsequent requests for the same task. So the thread itself runs only on a 1 second throttle now, always send first time requests, and waits 60 seconds for resending requests.

scale/job/execution/tasks/exe_task.py
```diff
@@ -86,15 +86,14 @@ def determine_error(self, task_update):
 
         raise NotImplementedError()
 
-    @abstractmethod
     def populate_job_exe_model(self, job_exe):
```
dfaller: The new pull task doesn't do anything to populate the model, so instead of forcing the pull task to define this method with no implementation that it doesn't care about, I made the method have a default no-op implementation that the pull task could ignore and the other tasks could overwrite as before.

PR #888, opened 2017-06-23
Issue 876 - Source File Data UI

### LINKED ISSUE #876: Source File Data UI
First a bug fix: switch the ingest detail UI page to use /v4/sources/{id}/. The v5 version does not include the product information.

Create some new pages for viewing source file data.

1. Add a new left-hand menu item called "Data".
2. "Data" should take you to a page where you can view source files in a table (eventually we'll have tabs for product files and general files as well). The source file table:
  a. should use the /v5/sources/ REST API
  b. should display the following columns: file name, file size, data started, data ended, countries, last modified
  c. allow filtering on file name
  d. allow datetime filtering with some way to allow the user to pick between last modified (call it processing time?) and data time
3. Clicking on a source file in the table should take you to a source file detail page.
  a. This page should use the /v5/sources/{id}/ detail REST API to display the general fields in a nice way.
  b. Have a download link for the file if the URL field is provided
  c. Using tabs, have three different tables for showing the ingests, jobs, and products related to the source file. These tables should use the /v5/sources/{id}/ingests/, /v5/sources/{id}/jobs/, and /v5/sources/{id}/products/ REST APIs. The job table can have the same filtering options as the normal job table. The products table should display file_name, file_size, job column with job type name/version text that links to the job detail page, data started, data ended, countries, last modified.

ericsvendsen-mil:
Closes #876

Here are the latest screenshots. The metadata "view all" button will only show when there are more than 9 metadata entries. 9 rows looks the best when compared to the details table.

I'm opening a separate issue for the UI back button problem - it's not limited to this particular UI view.

image

image

image


ericsvendsen-mil:
Done.
image

PR #896, opened 2017-06-29
Issue 894 - Recipe detail bug

### LINKED ISSUE #894: Recipe detail bug
The recipe detail page is attempting to retrieve information based off the wrong ID. Instead of doing another recipe type call, the recipe type should be built using the revision information included in the recipe details call.

ericsvendsen-mil:
Fixes #894 

The call to get the recipe type detail has been left in, and the resulting definition is being replaced with the revision definition. This was necessary due to the fact that job type data essential for the recipe viewer is not included in the recipe details response. So if the recipe type details call was not made, additional calls (potentially multiple) would have needed to be made further down the line to get the appropriate job type data.


ericsvendsen-mil:
Updated model to include necessary job information. As a result an extra call to retrieve recipe type details is no longer necessary.

PR #913, opened 2017-07-06
Issue 851

mikenholt:
* Refactored `recipe.models.RecipeManager.complete` to `complete_recipe` for uniformity
* Added `completed_recipe_count` and `completed_job_count` to `Batch` model 
* Added logic to complete_recipe to count as completed if part of a batch
* Added logic to complete_job to count as completed if part of a batch
* Added the two new fields to `BatchSerializer`
* Updated jobs


scale/batch/models.py
```diff
@@ -24,6 +24,26 @@
...
 
+    def count_completed_job(self, batch_id):
+        """Preforms a count-plus-one on the completed job field of a Batch
+
+        :param batch_id: The unique identifier of the batch.
+        :type batch_id: int
+        """
+
+        job_count = Batch.objects.filter(id=batch_id).values('completed_job_count')
+        job_count.update(completed_job_count=job_count[0]['completed_job_count'] + 1)
```
dfaller: You need these to be a single query to prevent race conditions. Look at Django models.F(). It will allow field + 1 queries.

Also "perform" is misspelled in the docstring.

scale/batch/models.py
```diff
@@ -24,6 +24,26 @@
...
+
+    def count_completed_recipe(self, batch_id):
+        """Preforms a count-plus-one on the completed recipe field of a Batch
+
+        :param batch_id: The unique identifier of the batch.
+        :type batch_id: int
+        """
+
+        recipe_count = Batch.objects.filter(id=batch_id).values('completed_recipe_count')
+        recipe_count.update(completed_recipe_count=recipe_count[0]['completed_recipe_count'] + 1)
```
dfaller: Same here.

Issue #901, opened 2017-06-30
Web docs css


ericsvendsen-mil:
Need to add the following somewhere in the sphinx pipeline to fix the text overflow and scrolling issue with the docs:

```
.wy-nav-content {
  max-width: none;
}

.wy-table-responsive table td, .wy-table-responsive table th {
  white-space: nowrap;
}

.wy-table-bordered-all, .rst-content table.docutils {
  width: 100%;
}
```

PR #919, opened 2017-07-07
#901 - Fix max width of docs

mikenholt:
Took the CSS provided by @ericsvendsen  and put it on a layout that extends all ReadTheDocs pages.


mikenholt:
Added additional styles to make the REST tables more manageable.  

DEMO: https://new-scale-mholt1.c9users.io/scale/docs/_build/html/rest/batch.html

Issue #816, opened 2017-04-18
Update REST API for settings secrets


dfaller:
Update the REST API to handle job configuration (settings/mounts) with secret values.

1. Create a validate() method for JobConfiguration that takes in a JobInterface, strips out any settings or mounts that are not in the interface, removes any settings that are secret, and returns a warning describing each setting/mount removed and why. See similar validate() methods for examples.
2. Add configuration to job type detail calls - GET /job-types/{id}/
Make sure to call JobConfiguration.validate() during job type detail calls. This will ensure that returned settings/mounts match any interface changes. Update REST API docs and add unit tests to ensure invalid/secret settings/mounts are not returned.
3. Add configuration to job type validation - POST /job-types/validation/
Make sure to call JobConfiguration.validate() and return warnings. Update REST API docs and add unit tests testing the validation.
4. Add configuration to job type edit - PATCH /job-types/{id}/
If there are secret settings passed in the configuration field, set these values in Vault. Make sure to call JobConfiguration.validate() before saving the JSON into the configuration field. Do calls to Vault after updating job type model, but within database transaction so Vault failure will roll back database changes. Update REST API docs and add unit tests (mock out Vault calls). Also please add the shared_memory_required field (this might just be a doc update).
5. Add configuration to job type import - POST /configuration/
Basically the same as (4), make them both go through the same code path (one way to edit a job type).
6. Add configuration to job type import validation - POST /configuration/validation/
Basically the same as (3), again, one code path
7. Add configuration to job type export - GET /configuration/
Basically the same as (2), one code path

PR #934, opened 2017-07-27
#816 - Update REST API for settings secrets

mikenholt:
1. Add `configuration` support to the following API endpoints:
    - GET `/job-types/{id}/`
    - POST `/job-types/validation/`
    - PATCH `/job-types/{id}/`
    - POST `/configuration/`
    - GET `/configuration/`
    - POST `/configuration/validation/` 
    - POST `/configuration/`

2. Updated `Vault` app - Added error to for when no secrets backend is configured but someone submits a secret to Scale.
3. Added two functions to job_config: `get_secret_settings` and `validate` - The first gathers the secrets before sending to `Vault` and the last checks for secrets and unused settings, as well as unused mounts, and discards them. 
4. Added function to JobType that will send secrets off to `Vault` 

**I will deploy this in our test_env to ensure it works with Vault.**

### DIFF
--- a/scale/job/views.py
+++ b/scale/job/views.py
@@ -87,6 +87,23 @@ class JobTypesView(ListCreateAPIView):
         except InvalidInterfaceDefinition as ex:
             raise BadParameter('Job type interface invalid: %s' % unicode(ex))
 
+        # Validate the job configuration and pull out secrets
+        configuration_dict = rest_util.parse_dict(request, 'configuration', required=False)
+        configuration = None
+        secrets = None
+        try:
+            if configuration_dict:
+                configuration = JobConfiguration(configuration_dict)
+                if interface:
+                    secrets = configuration.get_secret_settings(interface)
+                    configuration.validate(interface)
+                else:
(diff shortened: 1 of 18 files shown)


scale/job/views.py
```diff
@@ -200,6 +221,23 @@ def patch(self, request, job_type_id):
...
+        try:
+            if configuration_dict:
+                configuration = JobConfiguration(configuration_dict)
+                if interface:
+                    secrets = configuration.get_secret_settings(interface)
+                    configuration.validate(interface)
+                else:
+                    stored_interface = JobType.objects.values_list('interface', flat=True).get(pk=job_type_id)
+                    secrets = configuration.get_secret_settings(stored_interface)
+                    configuration.validate(JobInterface(stored_interface))
```
dfaller: These two lines can be pulled out of the if statement:
```
secrets = configuration.get_secret_settings(interface)
configuration.validate(interface)
```

Your if statement could simply check for `interface` and if None do the query to populate it.

PR #960, opened 2017-09-12
Issue 959 - Message Handler Service

dfaller:
Created a message handler service that launches long running message handlers to process the backend messages.

1. Created the concept of system services that launch long running system tasks
2. Created the message handler service and added ability to scheduler REST API to control the number of message handlers
3. Added node tasks and system tasks to status REST API
4. Added services section to status REST API


scale/job/tasks/manager.py
```diff
@@ -20,6 +22,50 @@ def __init__(self):
         self._tasks = {}  # {Task ID: Task}
         self._lock = threading.Lock()
 
+    def generate_status_json(self, nodes_list):
```
gisjedi: Didn't see any unit test coverage for this function.

scale/scheduler/tasks/services/messaging/messaging_service.py
```diff
@@ -0,0 +1,34 @@
...
+
+from scheduler.manager import scheduler_mgr
+from scheduler.tasks.services.messaging.message_handler_task import MessageHandlerTask
+from scheduler.tasks.services.service import Service
+
+
+logger = logging.getLogger(__name__)
+
+
+class MessagingService(Service):
```
gisjedi: Didn't see any unit test coverage for this class.

scale/scheduler/tasks/services/service.py
```diff
@@ -0,0 +1,126 @@
...
+import logging
+from abc import ABCMeta, abstractmethod
+
+from job.tasks.update import TaskStatusUpdate
+
+
+logger = logging.getLogger(__name__)
+
+
+class Service(object):
```
gisjedi: Didn't see any unit test coverage for this class.

scale/scheduler/tasks/system_task.py
```diff
@@ -46,3 +51,18 @@ def _add_database_docker_params(self):
                      DockerParameter('env', 'SCALE_DB_PORT=%s' % db['PORT'])]
 
         self._docker_params.extend(db_params)
+
+    def _add_messaging_docker_params(self):
```
gisjedi: Didn't see any unit test coverage for this function.

scale/job/tasks/manager.py
dfaller: Added unit test

scale/scheduler/tasks/services/messaging/messaging_service.py
dfaller: Unit tests for this class were added

scale/scheduler/tasks/services/service.py
dfaller: This is an abstract base class. It's concrete methods are tested in the unit tests for the messaging service.

scale/scheduler/tasks/system_task.py
dfaller: This method is tested within the unit tests for the messaging service. Being a very light method, I don't think a separate test for it is necessary.

dfaller:
I added needed unit tests, so I think we're good to go.

PR #961, opened 2017-09-13
Issue 953 - Scheduling fix

dfaller:
This PR is a scheduling fix that ignores queued jobs if the job references a job type or workspace that has not been synced to the scheduler yet. This can happen when a script loads things into Scale so quickly that the scheduler gets a job and doesn't know about its job type or workspaces yet.

### DIFF
--- a/scale/queue/job_exe.py
+++ b/scale/queue/job_exe.py
@@ -17,6 +17,7 @@ class QueuedJobExecution(object):
 
         self.id = queue.id
         self.is_canceled = queue.is_canceled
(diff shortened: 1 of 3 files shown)


dfaller:
It looks like I made this off of my other branch instead of master, so wait until the other gets merged before reviewing this one.

gisjedi:
FYI, I rebased this off master since I merged your previous PR. We've got a nice clean feature branch now and I'll merge once tests pass.

PR #964, opened 2017-09-19
Issue 962 - Use messaging system for canceled jobs on queue

dfaller:
This issue represents the first use of the new backend messaging system.

1. Created the first message type that bulk creates job_exe_end models.
2. Created a new scheduler background thread that regularly submits messages to the message broker.
3. Updated the scheduler to submit canceled job_exe_end instances from queued jobs to the new background thread (through the JobExecutionManager).


dfaller:
I still need to test this in Kiln.

Issue #656, opened 2016-12-06
Update recipe handling to recognize "changed" jobs


dfaller:
Update the recipe handling so that if deleted/new jobs in a recipe definition comparison share the same job_name, it is considered a "changed" job instead of a deleted + new job.

dfaller:
This has been merged.

PR #976, opened 2017-10-04
Issue 656 - Update recipe handling to recognize "changed" jobs

dfaller:
Updated recipe handling to detect changed jobs from one recipe definition to another.


dfaller:
Closes #656

Issue #954, opened 2017-09-07
REST APIs for files


dfaller:
We need some REST APIs for retrieving general files:

1. General file API at /files/
    - query params: started, ended, time_field (like from /products/), and file_name
    - move source_started and source_ended to the list of geospatial fields (making it apply to source files as well)
    - response should include all fields not specific to sources or products, including file_type

2. Job input file API at /jobs/{id}/input_files/
    - all inputs files for the given job, use the JobInputFile model, fall back to job_data if no JobInputFile models returned
    - same query params from (1.) plus job_input which filters on the job_input field in JobInputFile
    - response should be the same as (1.)

3. Recipe input file API at /recipes/{id}/input_files/
    - all inputs files for the given recipe, use the RecipeInputFile model, fall back to recipe_data if no RecipeInputFile models returned
    - same query params from (1.) plus recipe_input which filters on the recipe_input field in RecipeInputFile 
    - response should be the same as (1.)

All 3 APIs should use a general "filter_files" method on the ScaleFileManager.

PR #982, opened 2017-10-13
#954 - REST APIs for files

mikenholt:
* Added new REST API endpoints `/files/`, `/jobs/{id}/input_files/`, and `/recipes/{id}/input_files/`
* Added generic scale_file filter method that is used for each of the new endpoints.
* Filter params in include: 
  * `started`
  * `ended`
  * `time_field` (`data`, `source`, and `last_modified`)
  * `file_name`

* The `job` and `recipe` include a special filter based on the data for that object (`job_input`, `recipe_input`).  This includes logic to go back to the respective `*_data` definition if there is no `*InputFile` model entry.


scale/docs/rest/job.rst
```diff
@@ -1050,3 +1050,120 @@ These services provide access to information about "all", "currently running" an
...
++--------------------+-------------------+----------+---------------------------------------------------------------------+
+| time_field         | String            | Optional | The time series that the given time range is applied to.            |
+|                    |                   |          | Valid fields: 'source','data', and 'last_modified'                  |
++--------------------+-------------------+----------+---------------------------------------------------------------------+
+| file_name          | String            | Optional | The name of a specific file to be returned.                         |
++--------------------+-------------------+----------+---------------------------------------------------------------------+
+| job_input          | String            | Optional | The `input_data` entry name in the `job_data` JSON definition       |
+|                    |                   |          | of the job                                                          |
++--------------------+-------------------+----------+---------------------------------------------------------------------+
```
dfaller: Couple of tweaks to apply to all of the new API docs:
1. For time_field, list the three possible values and include a sentence describing what each one means. The sources API doc has an example of this.
2. For file_name, change the description to something like "Returns only input files with this file name".
3. For job_input/recipe_input, change the description to something like "Returns files for this job/recipe input".

scale/docs/rest/job.rst
```diff
@@ -1050,3 +1050,120 @@ These services provide access to information about "all", "currently running" an
...
+|        "workspace": {                                                                                                   |
+|            "id": 19,                                                                                                    |
+|            "name": "workspace-19"                                                                                       |
+|        },                                                                                                               |
+|        "countries": ["TCY", "TCT"],                                                                                     |
+|        "geometry" :null,                                                                                                |
+|        "center_point": null                                                                                             |
+|    }                                                                                                                    |
++-------------------------------------------------------------------------------------------------------------------------+
```
dfaller: This can return multiple files, so the docs should show an array with the paging fields. Applies to the other docs.

scale/job/models.py
```diff
@@ -1481,6 +1481,55 @@ class Meta(object):
...
+                job_input_file_ids = [f_id for f_id, name in job_input_files]
+
+        files = ScaleFile.objects.filter_files(started=started, ended=ended, time_field=time_field,
+                                               file_name=file_name)
+        files = files.filter(id__in=job_input_file_ids)
+        files = files.order_by('last_modified')
+
+        return files
+
+
```
dfaller: We would like to do this in one (or two for legacy) queries. So apply the filtering of job_id and job_input to the queryset return by ScaleFile.objects.filter_files() and see if any files are returned. If not, then basically do the logic in your else block.

scale/recipe/models.py
```diff
@@ -577,6 +577,57 @@ class Meta(object):
...
+                recipe_input_file_ids = [f_id for f_id, name in recipe_input_files if name == recipe_input]
+            else:
+                recipe_input_file_ids = [f_id for f_id, name in recipe_input_files]
+
+        files = ScaleFile.objects.filter_files(started=started, ended=ended, time_field=time_field,
+                                               file_name=file_name)
+        files = files.filter(id__in=recipe_input_file_ids)
+        files = files.order_by('last_modified')
+
+        return files
```
dfaller: Same as above, also there seems to be an extra space here.

scale/storage/models.py
```diff
@@ -217,6 +217,52 @@ def download_files(self, file_downloads):
...
+        files = ScaleFile.objects.all()
+        files = files.select_related('workspace')
+        files = files.defer('workspace__json_config')
+        files = files.prefetch_related('countries')
+
+        # Apply file_name filtering if present
+        if file_name:
+            files = files.filter(file_name=file_name)
+            files = files.order_by('last_modified')
+            return files
```
dfaller: File name could match more than one file, so don't stop short here. Continue applying the rest of the filtering.

scale/storage/serializers.py
```diff
@@ -117,4 +117,11 @@ class ScaleFileBaseSerializer(ModelIdSerializer):
 
 class ScaleFileSerializer(ScaleFileBaseSerializer):
     """Converts Scale file model fields to REST output"""
-    workspace = WorkspaceSerializer()
+
+    file_type = serializers.CharField()
+    data_type = serializers.CharField()
```
dfaller: Is this correct? data_type is defined differently in the ScaleFileBaseSerializer.

scale/storage/serializers.py
```diff
@@ -117,4 +117,11 @@ class ScaleFileBaseSerializer(ModelIdSerializer):
 
 class ScaleFileSerializer(ScaleFileBaseSerializer):
     """Converts Scale file model fields to REST output"""
-    workspace = WorkspaceSerializer()
+
+    file_type = serializers.CharField()
```
dfaller: Probably should go ahead and add file_type to the ScaleFileBaseSerializer.

scale/storage/serializers.py
```diff
@@ -117,4 +117,11 @@ class ScaleFileBaseSerializer(ModelIdSerializer):
...
+    data_type = serializers.CharField()
+    file_path = serializers.CharField()
+
+    source_started = serializers.DateTimeField()
+    source_ended = serializers.DateTimeField()
```
dfaller: Does anything use the base serializer? Otherwise the optional geospatial fields should probably be here instead of the base serializer.

scale/docs/rest/job.rst
```diff
@@ -1050,3 +1050,132 @@ These services provide access to information about "all", "currently running" an
...
+| ended              | ISO-8601 Datetime | Optional | The end of the time range to query.                                 |
+|                    |                   |          | Supports the ISO-8601 date/time format, (ex: 2015-01-01T00:00:00Z). |
+|                    |                   |          | Supports the ISO-8601 duration format, (ex: PT3H0M0S).              |
++--------------------+-------------------+----------+---------------------------------------------------------------------+
+| time_field         | String            | Optional | Indicates the time field(s) that *started* and *ended* will use for |
+|                    |                   |          | time filtering. Valid values are:                                   |
+|                    |                   |          |                                                                     |
+|                    |                   |          | - *last_modified* - last modification of source file meta-data      |
+|                    |                   |          | - *data* - data time of source file (*data_started*, *data_ended*)  |
```
dfaller: Should be *data* - data time of input file (*data_started*, *data_ended*)

dfaller: Applies to other docs too.

Issue #983, opened 2017-10-17
Create new job execution REST APIs


dfaller:
Create two new job executions REST APIs based on the new model structure of job_exe, job_exe_end, and job_exe_output.

1. /jobs/{job-id}/executions/ should return a paged list of job executions ordered by execution number descending. Each job execution should contain all fields (except for JSON fields) from job_exe and job_exe_end. If a job_exe model does not have a corresponding job_exe_end model, then its status should be 'RUNNING'.

2. /jobs/{job-id}/executions/{exe-num}/ should return the details for the job execution with the given job ID and execution number. It should include all fields from the job_exe, job_exe_end, and job_exe_output models.

3. Deprecate the /jobs/executions/, /job-executions/ and /job-executions/{id}/ APIs (if v6, return 404, mark the views and supporting classes/serializers as deprecated). The only execution API to not touch right now is the logging one: job-executions/{id}/logs/(stdout|stderr|combined)/.

PR #1017, opened 2017-12-14
#983 - Create new job execution REST APIs

mikenholt:
* Deprecated old views, functions, tests, serializers, and docs. 
 * Added 2 new `/jobs/` endpoints for viewing job_exes associated with the requested job_id.
 * Added views, functions, and serializers to handle the two new endpoints.

### DIFF
--- a/scale/docs/rest/job.rst
+++ b/scale/docs/rest/job.rst
@@ -1179,3 +1179,203 @@ These services provide access to information about "all", "currently running" an
 |        ]                                                                                                                |
 |    }                                                                                                                    |
 +-------------------------------------------------------------------------------------------------------------------------+
+
+.. _rest_job_execution_list:
+
++-------------------------------------------------------------------------------------------------------------------------+
+| **Job Executions List**                                                                                                 |
++=========================================================================================================================+
+| Returns detailed information about input files associated with a given Job ID.                                          |
++-------------------------------------------------------------------------------------------------------------------------+
+| **GET** /jobs/{id}/executions/                                                                                          |
+|         Where {id} is the unique identifier of an existing job.                                                         |
++-------------------------------------------------------------------------------------------------------------------------+
+| **Query Parameters**                                                                                                    |
++--------------------+-------------------+----------+---------------------------------------------------------------------+
+| started            | ISO-8601 Datetime | Optional | The start of the time range to query.                               |
+|                    |                   |          | Supports the ISO-8601 date/time format, (ex: 2015-01-01T00:00:00Z). |
+|                    |                   |          | Supports the ISO-8601 duration format, (ex: PT3H0M0S).              |
++--------------------+-------------------+----------+---------------------------------------------------------------------+
+| ended              | ISO-8601 Datetime | Optional | The end of the time range to query.                                 |
+|                    |                   |          | Supports the ISO-8601 date/time format, (ex: 2015-01-01T00:00:00Z). |
+|                    |                   |          | Supports the ISO-8601 duration format, (ex: PT3H0M0S).              |
++--------------------+-------------------+----------+---------------------------------------------------------------------+
+| status             | String            | Optional | Return only executions with a status matching this string.          |
+|                    |                   |          | Choices: [QUEUED, RUNNING, FAILED, COMPLETED, CANCELED].            |
+|                    |                   |          | Duplicate it to filter by multiple values.                          |
++--------------------+-------------------+----------+---------------------------------------------------------------------+
+| node_id            | Integer           | Optional | Return only executions that ran on a given node.                    |
+|                    |                   |          | Duplicate it to filter by multiple values.                          |
++--------------------+-------------------+----------+---------------------------------------------------------------------+
+| **Successful Response**                                                                                                 |
++--------------------+----------------------------------------------------------------------------------------------------+
+| **Status**         | 200 OK                                                                                             |
++--------------------+----------------------------------------------------------------------------------------------------+
+| **Content Type**   | *application/json*                                                                                 |
++--------------------+----------------------------------------------------------------------------------------------------+
+| **JSON Fields**                                                                                                         |
++--------------------+-------------------+--------------------------------------------------------------------------------+
+| count              | Integer           | The total number of results that match the query parameters.                   |
++--------------------+-------------------+--------------------------------------------------------------------------------+
+| next               | URL               | A URL to the next page of results.                                             |
++--------------------+-------------------+--------------------------------------------------------------------------------+
+| previous           | URL               | A URL to the previous page of results.                                         |
++--------------------+-------------------+--------------------------------------------------------------------------------+
+| results            | Array             | List of result JSON objects that match the query parameters.                   |
++--------------------+-------------------+--------------------------------------------------------------------------------+
+| .id                | Integer           | The unique identifier of the model. Can be passed to the details API call.     |
+|                    |                   | (See :ref:`Job Execution Details <rest_job_execution_details>`)                |
++--------------------+-------------------+--------------------------------------------------------------------------------+
+| .status            | String            | The status of the job execution.                                               |
+|                    |                   | Choices: [QUEUED, RUNNING, FAILED, COMPLETED, CANCELED].                       |
++--------------------+-------------------+--------------------------------------------------------------------------------+
+| .command_arguments | String            | The argument string to execute on the command line for this job execution.     | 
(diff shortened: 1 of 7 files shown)


scale/job/models.py
```diff
@@ -1049,6 +1051,77 @@ def get_exes(self, started=None, ended=None, statuses=None, job_type_ids=None, j
...
+        :rtype: [:class:`job.models.JobExecution`]
+        """
+
+        # Fetch a list of job executions
+        job_exe = JobExecution.objects.all().select_related('job', 'job_type', 'node', 'jobexecutionend',
+                                                            'jobexecutionoutput')
+        job_exe = job_exe.defer('stdout', 'stderr')
+
+        # Apply job and execution filtering
+        job_exe = job_exe.filter(job__id=job_id, exe_num=exe_num)
```
dfaller: I would recommend using get() instead of filter() to guarantee that only one row matches.

scale/job/views.py
```diff
@@ -764,7 +778,43 @@ class JobExecutionsView(ListAPIView):
...
+            ended = rest_util.parse_timestamp(request, 'ended', required=False)
+            rest_util.check_time_range(started, ended)
+
+            statuses = rest_util.parse_string_list(request, 'status', required=False)
+            node_ids = rest_util.parse_int_list(request, 'node_id', required=False)
+
+            job_exes = JobExecution.objects.get_job_exes(job_id=job_id, started=started, ended=ended,
+                                                         statuses=statuses, node_ids=node_ids)
+
+            if job_exes.exists():
```
dfaller: exists() does a separate query, so we don't want this. Just pass job_exes to the paginator.

scale/job/views.py
```diff
@@ -798,13 +848,43 @@ class JobExecutionDetailsView(RetrieveAPIView):
...
+        if not exe_num:
+            if request.version == 'v5':
+                job_exe_id = job_id
+                return self.retrieve_v5(request, job_exe_id)
+            else:
+                raise Http404
+        else:
+            job_exe = JobExecution.objects.get_job_exe_details(job_id=job_id, exe_num=exe_num)
+
+            if job_exe.exists():
```
dfaller: exists() does a separate query, so we don't want this. Use this pattern for when trying to retrieve one specific model with details.

```
try:
    job_exe = JobExecution.objects.get_logs(job_exe_id)
except JobExecution.DoesNotExist:
    raise Http404
```

This pattern should be combined with using the get() method in detail retrieval on the model manager because the get() method will throw an exception if there is not exactly one row that matches.

scale/docs/rest/job.rst
```diff
@@ -1179,3 +1179,369 @@ These services provide access to information about "all", "currently running" an
...
+|          "version": "1.1",                                                                                                |
+|          "errors": [],                                                                                                    |
+|          "parse_results": []                                                                                              |
+|      },                                                                                                                   |
+|      "configuration": {                                                                                                   |
+|          "tasks": [...],                                                                                                  |
+|          "version": "2.0"}                                                                                                |
+|  }                                                                                                                        |
++---------------------------------------------------------------------------------------------------------------------------+
```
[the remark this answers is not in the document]
mikenholt: Do you want the more details API return to still include base/model serializers?

scale/docs/rest/job.rst
```diff
@@ -1179,3 +1179,369 @@ These services provide access to information about "all", "currently running" an
...
+|                "cpus_scheduled": 1.0,                                                                                     |
+|                "mem_scheduled": 256,                                                                                      |
+|                "disk_out_scheduled": 10.0,                                                                                |
+|                "disk_in_scheduled": 10.0                                                                                  |
+|            },                                                                                                             |
+|            ...                                                                                                            |
+|        ]                                                                                                                  |
+|    }                                                                                                                      |
++---------------------------------------------------------------------------------------------------------------------------+
```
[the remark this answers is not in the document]
mikenholt: Should the job details (`/jobs/{id}/` - [spec](http://ngageoint.github.io/scale/docs/rest/job.html#rest-job-details)) return the old or new set of base fields?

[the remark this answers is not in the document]
dfaller: The old set of base fields. The next issue is to create a new v6 version of job details.

scale/docs/rest/job.rst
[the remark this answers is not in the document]
dfaller: I think the base serializers will be sufficient.

scale/docs/rest/job.rst
```diff
@@ -1179,3 +1179,260 @@ These services provide access to information about "all", "currently running" an
...
+| **JSON Fields**                                                                                                           |
++----------------------+-------------------+--------------------------------------------------------------------------------+
+| id                   | Integer           | The unique identifier of the model. Can be passed to the details API call.     |
+|                      |                   | (See :ref:`Job Execution Details <rest_job_execution_details>`)                |
++----------------------+-------------------+--------------------------------------------------------------------------------+
+| status               | String            | The status of the job execution.                                               |
+|                      |                   | Choices: [RUNNING, FAILED, COMPLETED, CANCELED].                               |
++----------------------+-------------------+--------------------------------------------------------------------------------+
+| command_arguments    | String            | The argument string to execute on the command line for this job execution.     |
```
dfaller: 1. command_arguments and environment should be removed (in the field listing and in the example JSON)
2. exe_num and cluster_id should be added to the example JSON and field listing
3. results_manifest should be removed from both the field listing and example JSON
4. results should be renamed output for both field listing and example JSON with a description saying its a JSON describing the job output. There should be a JSON schema for this to link to.
5. The link for configuration should point to the execution configuration, not the job configuration.
6. The node section should not have port and slave_id.
7. task_results needs to be added to field listing and example JSON.
8. last_modified should be removed (in the field listing and in the example JSON)
9. Re-order fields to match the serializer (the same order that will be returned for real)

scale/docs/rest/job.rst
```diff
@@ -1179,3 +1179,260 @@ These services provide access to information about "all", "currently running" an
...
+|    {                                                                                                                      |
+|        "count": 57,                                                                                                       |
+|        "next": null,                                                                                                      |
+|        "previous": null,                                                                                                  |
+|        "results": [                                                                                                       |
+|            {                                                                                                              |
+|                "id": 3,                                                                                                   |
+|                "status": "COMPLETED",                                                                                     |
+|                "command_arguments": "",                                                                                   |
```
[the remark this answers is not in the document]
dfaller: last_modified should be removed (in the field listing and in the example JSON)
Re-order fields to match the serializer (the same order that will be returned for real)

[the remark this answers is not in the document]
dfaller: These same changes need applied to the serializers.

scale/docs/rest/job.rst
dfaller: These same changes need applied to the serializers.

scale/job/serializers.py
```diff
@@ -281,25 +298,41 @@ class JobExecutionSerializer(JobExecutionBaseSerializer):
...
-    disk_in_scheduled = serializers.FloatField(source='input_file_size')
-    disk_out_scheduled = serializers.FloatField()
-    disk_total_scheduled = serializers.FloatField()
 
-    results = serializers.JSONField(default=dict, source='jobexecutionoutput.output')
+class JobExecutionDetailsSerializer(JobExecutionSerializer):
+    """Converts job execution model fields to REST output"""
 
-    results_manifest = serializers.JSONField(default=dict)
+    task_results = serializers.JSONField(default=dict)
```
dfaller: task_results will need its source to point to the JobExecutionEnd model.

Also I think it would be better if the JSON fields default to null (like error model above) instead of being an empty dict.

scale/job/serializers.py
```diff
@@ -198,6 +198,21 @@ class JobExecutionBaseSerializer(ModelIdSerializer):
...
+    timeout = serializers.IntegerField()
+
+    created = serializers.DateTimeField()
+    queued = serializers.DateTimeField()
+    started = serializers.DateTimeField()
+    ended = serializers.DateTimeField(source='jobexecutionend.ended')
+
+    job = ModelIdSerializer()
+    node = ModelIdSerializer()
+    error = ModelIdSerializer()
```
dfaller: Add job_type = ModelIdSerializer() to base serializer
Move timeout from JobExecutionBaseSerializer to JobExecutionSerializer
Move exe_num and cluster_id from JobExecutionSerializer to JobExecutionBaseSerializer
Update docs appropriately

PR #1062, opened 2018-01-02
Issue 394- Configurable job debug logging

### LINKED ISSUE #394: Configurable job debug logging
Implement a configurable switch for logging level that can be set that gets passed to all tasks that run within the Scale Docker image (system jobs and pre/post tasks). The configured log level can be passed to the tasks via environment variables that set the appropriate logging level in the local_settings.py. The configured logging level can be stored in the scheduler model with a default value of INFO, allowing the value to be configurable via the REST API and UI.

Fizz11:
added logging_level to scheduler, will allow debugging level to be set through rest API

1. Added logging_level property to scheduler model with default value of INFO
2. updated configurators.py to pass logging_level to _configure_all_tasks and set a new docker parameter LOGGING_LEVEL to the passed in value
3. updated settings and local_settings_docker.py to set loging level appropriately based on the docker parameter
4. fixed a small bug in local_docker_settings.py where a call to .lower() was missing the parentheses. 
5. updated tests to account for new docker parameter
6. updated views.py to allow updating field from REST API

fixes #394


dfaller:
@gisjedi Yes, there is value in separating `SCALE_DEBUG` and `LOGGING_LEVEL`. With `SCALE_DEBUG` you are running Django in a debug mode that is not suitable/secure for production use. This has more substantial consequences than just raising the level of logging, so I would keep Django's debug flag separate from the logging level of our components.

scale/job/configuration/configurators.py
```diff
@@ -234,7 +234,7 @@ def __init__(self, workspaces):
                                  'SCALE_DB_PORT': db['PORT']}
         self._system_settings_hidden = {key: '*****' for key in self._system_settings.keys()}
 
-    def configure_scheduled_job(self, job_exe, job_type, interface):
+    def configure_scheduled_job(self, job_exe, job_type, interface, debug_level):
```
dfaller: I would rename `debug_level` to `logging_level`. Also be sure to add the new parameter to the docstring.

scale/job/configuration/configurators.py
```diff
@@ -250,7 +250,7 @@ def configure_scheduled_job(self, job_exe, job_type, interface):
         :returns: A copy of the configuration containing secret values
         :rtype: :class:`job.configuration.json.execution.exe_config.ExecutionConfiguration`
         """
-
+
```
dfaller: Remove extra whitespace

scale/job/configuration/configurators.py
```diff
@@ -263,15 +263,15 @@ def configure_scheduled_job(self, job_exe, job_type, interface):
...
+        self._configure_all_tasks(config, job_exe, job_type, debug_level)
 
         # Configure secrets
         config_with_secrets = self._configure_secrets(config, job_exe, job_type, interface)
 
         job_exe.configuration = config.get_dict()
         return config_with_secrets
 
-    def _configure_all_tasks(self, config, job_exe, job_type):
+    def _configure_all_tasks(self, config, job_exe, job_type, debug_level):
```
dfaller: I would rename `debug_level` to `logging_level`. Also be sure to add the new parameter to the docstring.

scale/job/configuration/configurators.py
```diff
@@ -318,14 +318,15 @@ def _configure_all_tasks(self, config, job_exe, job_type):
         # Configure tasks for logging
         if settings.LOGGING_ADDRESS is not None:
             log_driver = DockerParameter('log-driver', 'syslog')
+            debug_level_parameter = DockerParameter('env','LOGGING_LEVEL=%s' % debug_level)
```
dfaller: Instead of directly adding a Docker parameter here, use an env var. They will be converted correctly to appropriate Docker parameters later in the configurator.

```
logging_env_vars = {'LOGGING_LEVEL': logging_level}
config.add_to_task('pre', env_vars=logging_env_vars, docker_params=[...])
```

scale/scheduler/configuration.py
```diff
@@ -25,3 +25,4 @@ def __init__(self, scheduler=None):
             self.is_paused = scheduler.is_paused
             self.num_message_handlers = scheduler.num_message_handlers
             self.queue_mode = scheduler.queue_mode
+            self.logging_level = scheduler.logging_level
```
dfaller: `self.logging_level` needs to be set with a default value like the others.

scale/scale/settings.py
```diff
@@ -285,7 +291,7 @@
     'handlers': LOG_HANDLERS,
     'loggers': {
         '': {
-            'handlers': ['console', 'console-err'],
+            'handlers': ['console-err-debug'],
```
dfaller: Why are we forcing all log levels through stderr?

Fizz11: @gisjedi and I discussed it, couldn't think of a good reason for logging to show up on more than one output... 
I'm OK changing it back if you think it otherwise

gisjedi: In the long distance past we had separate files for error and above and then a full log file. I'm now of the opinion log output is best sent in its entirety to STDERR... STDOUT is more intended for *nix process data piping to downstream STDIN. In any case it's all going to end up in Elasticsearch in the future where it can be uniformly queried and filtered as desired.

gisjedi:
@dfaller after chatting about the confusing a bit more offline, @Fizz11 and I came to the conclusion that for future purposes we'd potentially want to configure system components (Web Server, Scheduler, other Marathon managed components) consistently apart from scheduled managed tasks (Scan, Strike, Batch, Pre, Post, Message Handlers, etc.). Our final determination was to have two environment variables supported within Scale image: `SYSTEM_LOGGING_LEVEL` and `JOB_LOGGING_LEVEL`. 

The application of them is given below:

* `SYSTEM_LOGGING_LEVEL`: Applied to scheduler and pushed via `bootstrap.py` to all Marathon launched components. This would supercede the use of `SCALE_DEBUG`. Valid values would be consistent with `JOB_LOGGING_LEVEL`. The one addition would be that setting DEBUG would put Django in DEBUG mode as well.
* `JOB_LOGGING_LEVEL`: By convention (to be documented in new Sphinx Architecture docs), only applied to tasks launched by Scale Scheduler via Mesos. This environment variable will be renamed from `LOGGING_LEVEL` as implemented presently. This would be obviously supported for Scale components by our use in `settings.py`, but would allow algorithm developers to consume for their use in tweaking log levels as well.

@dfaller Feel free to disagree on this approach, but that seemed the most consistent to me.

dfaller:
@gisjedi I'm fine with splitting the logging level into two for system processes and jobs, though I would handle all Scale system jobs under the `SYSTEM_LOGGING_LEVEL`.

The `JOB_LOGGING_LEVEL` should be named appropriately and added to the Seed specification. Also, this value should be configurable per job type, not one value at the system/scheduler level. I think it should be its own issue and will require integration with the job configuration specification.

I would still separate DEBUG and INFO on stdout and WARNING and above on stderr.

gisjedi:
Tracking Seed spec update https://github.com/ngageoint/seed/issues/96

Fizz11:
1. changed SCALE_DEBUG to DJANGO_DEBUG
2. SYSTEM_LOGGING_LEVEL will effect logging level of marathon launched components. 
3. system_job_logging_level is set through REST API and only effects system level jobs. 
4. updated tests and documentation

scale/scale/settings.py
```diff
@@ -245,11 +248,18 @@
         'stream': sys.stdout
     },
     'console-err': {
-        'level': 'WARNING',
+        'level': 'DEBUG',
         'class': 'logging.StreamHandler',
         'formatter': 'standard',
         'stream': sys.stderr
     },
+    'console-debug': {
```
gisjedi: I'd call this `console-out` or `console-stdout` and change `console-err` to `console-stderr`. Pick your consistent poison. :-) `console-debug` just doesn't indicate its target output to me.

scale/scale/settings.py
```diff
@@ -245,11 +248,18 @@
         'stream': sys.stdout
     },
     'console-err': {
```
gisjedi: Don't we want to leave this at `WARNING` so we aren't duplicating messages between `stderr` and `stdout`?

scale/scheduler/models.py
```diff
@@ -91,6 +91,7 @@ def get_status(self):
             'is_online': False,
             'is_paused': False,
             'hostname': None,
+            'system_job_logging_level':''
```
gisjedi: Shouldn't we have the default of `INFO` here?

scale/scale/settings.py
Fizz11: right now all logging messages go to stderr and info+debug go to stdout. all logging info should definitely show up on stderr, so you have one complete set of logs in a single stream when debugging. info_debug doesn't necessarily have to be on stdout.

gisjedi: I definitely agree with using `stderr` for everything, but the goal of using `stdout` and `stderr` is for ease of delineation between the streams. Ultimately both streams are interleaved in the Scale log viewer, so we are just going to get duplicates of `INFO` and `DEBUG` levels as currently implemented. I think this should be revisited in the future, if we can support colorization of logs from a single stream based on level. I'd recommend putting in a UI issue for this.

scale/job/configuration/configurators.py
```diff
@@ -317,7 +317,7 @@ def _configure_all_tasks(self, config, job_exe, job_type):
 
         # Configure tasks for logging
         if settings.LOGGING_ADDRESS is not None:
-            log_driver = DockerParameter('log-driver', 'syslog')
+            log_driver = DockerParameter('log-driver', 'syslog')
```
dfaller: Remove extra whitespace

scale/docs/rest/scheduler.rst
```diff
@@ -23,18 +23,22 @@ There should be one and only one scheduler entry in the database and it is used
...
-+----------------------+-------------------+------------------------------------------------------------------------------+
++--------------------------+---------------+------------------------------------------------------------------------------+
+| is_paused                | Boolean       | True if the scheduler is paused. This functions like individually pausing    |
+|                          |               | all nodes but maintains separated state so toggling this back to unpaused    |
+|                          |               | results in the previous individual node pause state.                         |
++--------------------------+---------------+------------------------------------------------------------------------------+
+| num_message_handlers     | Integer       | The number of message handlers to have scheduled                             |
++--------------------------+---------------+------------------------------------------------------------------------------+
+| system_job_logging_level | String        | The logging level sent to all system tasks that run within                   |
```
dfaller: Rename `system_job_logging_level` to `system_logging_level` and update the description please. This setting is for all Scale system components/pieces, not just jobs.

scale/scheduler/models.py
```diff
@@ -80,7 +80,7 @@ def get_status(self):
         """Fetch summary hardware resource usage for the scheduler framework.
 
         :returns: Node resource usage information.
-        :rtype: dict
+        :rtype: dict
```
dfaller: Remove extra whitespace

scale/scale/local_settings_docker.py
```diff
@@ -104,7 +104,18 @@
 CONFIG_URI = os.environ.get('CONFIG_URI', CONFIG_URI)
 
 # Logging configuration
-LOGGING = LOG_CONSOLE_DEBUG if DEBUG else LOG_CONSOLE_INFO
+LOGGING_LEVEL = os.environ.get('SYSTEM_LOGGING_LEVEL', '') #should only be present in running jobs
```
dfaller: I would remove the comment since that env var will be used for other Scale components, not just jobs.

Issue #985, opened 2017-10-17
New v6 recipe details REST API


dfaller:
Deprecate the recipe details API and create a new v6 version that removes the "inputs" section, rename *data* to *input* (change both recipe model and serializer) and look at changing the recipe_type serializer so that the recipe_type definition is not included (so the only definition comes from recipe_type_rev).

Make sure the v5 version works the same as before.

PR #1095, opened 2018-01-23
#985 - New v6 recipe details REST API

mikenholt:
- removed `inputs` section and associated logic from `get_details` in `models.py`
- renamed `data` to `input` in recipe model
- downgraded `recipe_type` serialized level for `RecipeDetailsSerializer` (relying on `RecipeBaseSerializer` now)

These changes were put in place for `v6` requests to `/recipes/{id}`, `/recipes/{id}/reprocess/ `, and `/queue/new-recipe/`


scale/queue/models.py
```diff
@@ -398,7 +398,7 @@ def queue_new_recipe(self, recipe_type, data, event, superseded_recipe=None, del
         :param recipe_type: The type of the new recipe to create
         :type recipe_type: :class:`recipe.models.RecipeType`
         :param data: The recipe data to run on, should be None if superseded_recipe is provided
-        :type data: :class:`recipe.data.recipe_data.RecipeData`
+        :type data: :class:`recipe.input.recipe_data.RecipeData`
```
dfaller: It looks like this shouldn't have been changed since the directory is still named `data`.

scale/queue/models.py
```diff
@@ -444,7 +444,7 @@ def queue_new_recipe_for_user(self, recipe_type, data):
...
```
dfaller: Same here.

scale/recipe/models.py
```diff
@@ -59,8 +59,8 @@ def create_recipe(self, recipe_type, data, event, superseded_recipe=None, delta=
 
         :param recipe_type: The type of the recipe to create
         :type recipe_type: :class:`recipe.models.RecipeType`
-        :param data: The recipe data to run on, should be None if superseded_recipe is provided
-        :type data: :class:`recipe.data.recipe_data.RecipeData`
+        :param input: The recipe input to run on, should be None if superseded_recipe is provided
+        :type input: :class:`recipe.input.recipe_data.RecipeData`
```
dfaller: Same here.

PR #1097, opened 2018-01-24
Issue 1073 - Add Scale info to job env vars

dfaller:
1. Added a nullable batch_id field to recipe model
2. Added nullable batch_id, recipe_id, and root_recipe_id to job model
3. Added nullable batch_id and recipe_id to queue model
4. Added nullable batch_id and recipe_id to job_exe model
5. When a job execution is configured, the job_exe model has the job, recipe, and batch IDs to insert as env vars.
6. Added logic to the database update system task to populate the new fields on the job and recipe models.


dfaller:
They could be, but I was reluctant to redo the migrations and my database.

PR #1111, opened 2018-02-06
Issue 183 - Refactor Job Requeue

dfaller:
I created two new messages to handle re-queuing jobs and created a new re-queue jobs API (in the jobs app) that sends a message to perform the re-queue and quickly responds. I deprecated the old re-queue jobs API (in the queue app).

### DIFF
--- a/scale/queue/messages/requeue_jobs_bulk.py
+++ b/scale/queue/messages/requeue_jobs_bulk.py
@@ -55,7 +55,7 @@ def create_requeue_jobs_bulk_message(started=None, ended=None, error_categories=
 
 
 class RequeueJobsBulk(CommandMessage):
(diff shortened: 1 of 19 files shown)


scale/queue/messages/requeue_jobs_bulk.py
```diff
@@ -0,0 +1,170 @@
...
+    message.job_ids = job_ids
+    message.job_type_ids = job_type_ids
+    message.priority = priority
+    message.status = status
+
+    return message
+
+
+class RequeueJobsBulk(CommandMessage):
+    """Command message that sets performs a bulk re-queue operation
```
mikenholt: Wording is weird.

PR #1127, opened 2018-03-14
Issue 1037 - Batch Metrics

dfaller:
Created a new message that calculates metrics for a batch, including metrics per job name (node within a recipe graph).

New metrics fields were added to the batch model and a new model BatchMetrics was created. New fields are storing the Seed main task run time were also added to the job_exe_end model.


scale/batch/messages/update_batch_metrics.py
```diff
@@ -0,0 +1,94 @@
...
+            message = UpdateBatchMetrics()
+        message.add_batch(batch_id)
+    if message:
+        messages.append(message)
+
+    return messages
+
+
+class UpdateBatchMetrics(CommandMessage):
+    """Command message that updates recipe metrics
```
mikenholt: "Command message that updates **batch** metrics"?

Issue #1117, opened 2018-02-19
Create a system job that will unpublish a file 


mikenholt:
Scale needs the ability to spin up a system job that can unpublish a file.  Additionally, this job also needs to have an option that can also delete or purge the file.

PR #1130, opened 2018-03-22
#1117 - Create a system job that will unpublish a file

mikenholt:
- Added `scale_destroy_file`  system command 
- Added they system job definition 
- Integrated a message to modify `ScaleFile`


scale/storage/fixtures/destroy_file_errors.json
```diff
@@ -0,0 +1,15 @@
+[
+    {
+        "model": "error.Error",
```
dfaller: I don't think we need this error. If the file doesn't exist, then we can consider the job successful.

scale/storage/fixtures/destroy_file_job_type.json
```diff
@@ -0,0 +1,47 @@
+[
+	{
+		"model": "job.JobType",
+		"pk": null,
+		"fields": {
+			"name": "scale-destroy-file",
```
dfaller: I would rename the job to `scale-delete-file` instead of `destroy`.

scale/storage/management/commands/scale_destroy_file.py
```diff
@@ -0,0 +1,63 @@
...
+        file_ids = options.get('file_ids')
+        job_id = options.get('job_id')
+
+        logger.info('Command starting: scale_destroy_file')
+        logger.info('File IDs: %s', file_ids)
+        logger.info('Job ID: %i', job_id)
+
+        file_info = ScaleFile.objects.filter(id__in=file_ids).values('id', 'file_path')
+        for f in file_info:
+            destroy_file_job.destroy_file(f['file_path'], job_id)
```
dfaller: The actual job code can't just do a simple os.remove(). It needs to use the workspace brokers that know how to handle the files in a given storage environment. For instance, deleting off of S3 is going to require specific S3 calls done in boto. Also there will need to be code added to the configuration logic so that job is provided with the necessary workspace mounts.

scale/storage/management/commands/scale_destroy_file.py
```diff
@@ -0,0 +1,63 @@
...
+
+        messages = create_delete_files_messages(file_ids=file_ids, purge=True)
```
dfaller: This job should be able to handle both deleting and purging files.

scale/storage/management/commands/scale_destroy_file.py
```diff
@@ -0,0 +1,63 @@
...
+        """
+
+        # Register a listener to handle clean shutdowns
+        signal.signal(signal.SIGTERM, self._onsigterm)
+
+        file_ids = options.get('file_ids')
+        job_id = options.get('job_id')
+
+        logger.info('Command starting: scale_destroy_file')
+        logger.info('File IDs: %s', file_ids)
```
dfaller: We want to make this job (and other new system jobs) not rely on a Scale database connection. We need to pass in all of the meta-data we need for deleting the files, such as what workspace/broker they exist on, file ID, path within the workspace, etc.

scale/storage/messages/delete_files.py
```diff
@@ -0,0 +1,109 @@
...
+        """
+
+        with transaction.atomic():
+            if self.purge:
+                ScaleFile.objects.filter(id__in=self._file_ids).delete()
+            else:
+                scale_files = ScaleFile.objects.filter(id__in=self._file_ids)
+                for scale_file in scale_files:
+                    scale_file.set_deleted()
+                    scale_file.save()
```
dfaller: Setting all of the files to deleted should be done in a single database query. Once this is done, there is no need for the atomic transaction.

scale/storage/management/commands/scale_destroy_file.py
```diff
@@ -0,0 +1,85 @@
...
+    def add_arguments(self, parser):
+        parser.add_argument('-f', '--files', nargs='+', type=json.loads, required=True,
+                            help='File path and ID in a JSON string.' +
+                            ' e.g: "{"file_path":"some.file", "id":"399"}"')
+        parser.add_argument('-j', '--job_id', action='store', type=int, required=True,
+                            help='ID of the Job model')
+        parser.add_argument('-w', '--workspace', action='store', type=json.loads, required=True,
+                            help='Workspace configuration in a JSON string.')
+        parser.add_argument('-p', '--purge', action='store', type=bool, required=True,
+                            help='Purge all records for the given file')
```
dfaller: I would recommend using env vars for the inputs instead of the command line. There could be any number of workspaces that house the files, so you will need to be able to specify multiple workspaces and indicate which workspace each file belongs to. I would also recommend trying a strategy like this to break up the inputs:
NUM_FILES = 5
FILE_1_ID = 1
FILE_1_PATH = relative/workspace/path/file.txt
FILE_1_WORKSPACE = workspace_1
...
FILE_5_ID = 5
FILE_5_PATH = relative/workspace/path/file_5.txt
FILE_5_WORKSPACE = workspace_2
NUM_WORKSPACES = 2
WORKSPACE_1_NAME = workspace_1
WORKSPACE_1_BROKER = {"json": "config"}
WORKSPACE_2_NAME = workspace_2
WORKSPACE_2_BROKER = {"json": "config"}

scale/storage/management/commands/scale_destroy_file.py
```diff
@@ -0,0 +1,85 @@
...
+                                                             volume_path=workspace_config['broker']['host_path'],
+                                                             broker=broker)
+
+        if destroy_job_return == 0:
+            messages = create_delete_files_messages(files=files, purge=purge)
+            CommandMessageManager().send_messages(messages)
+
+            logger.info('Command completed: scale_destroy_file')
+
+        return destroy_job_return
```
dfaller: There's a way we define common system exit codes for system errors. Take a look at pre and post tasks for examples. It prevents the magic numbers used here.

scale/storage/messages/delete_files.py
```diff
@@ -0,0 +1,104 @@
...
+        return message
+
+    def execute(self):
+        """See :meth:`messaging.messages.message.CommandMessage.execute`
+        """
+
+        if self.purge:
+            ScaleFile.objects.filter(id__in=self._file_ids).delete()
+        else:
+            ScaleFile.objects.filter(id__in=self._file_ids).update(is_deleted=False, deleted=timezone.now())
```
dfaller: We should probably turn this into a manager method. We also need to consider unpublishing the files. If a file is deleted, it needs to be unpublished. We already have an unpublish_job message, but we'll have to think through how this will all work together.

scale/storage/messages/delete_files.py
```diff
@@ -0,0 +1,104 @@
...
+            messages.append(message)
+            message = DeleteFiles()
+            message.purge = purge
+        message.add_file(scale_file.id)
+    if message:
+        messages.append(message)
+
+    return messages
+
+class DeleteFiles(CommandMessage):
```
dfaller: Make sure to add this message to the Wiki.

dfaller:
The key class for system job error exceptions is error.exceptions.ScaleError. It connects together an exception that can be thrown with a builtin system error. The pre and post tasks have an example of the `except` code that handles ScaleErrors thrown at the top level in their commands. If you're using an already existing builtin system error, there might already be a ScaleError for you to use.

PR #1151, opened 2018-04-24
Bug fix for Issue 670

dfaller:
Bug fix for updating old batches with recipe_type_rev_id

### DIFF
--- a/scale/scheduler/database/updater.py
+++ b/scale/scheduler/database/updater.py
@@ -153,11 +153,13 @@ class DatabaseUpdater(object):
             cursor.execute(qry_3, [str(batch_id), str(batch_id), str(batch_id)])
             count = cursor.rowcount
             if count:
-                logger.info('Batch with batch_id %d set to correct recipe type revision', batch_id)
+                logger.info('Batch with batch_id %d set to correct recipe type revision (based on old batch recipe)',
+                            batch_id)
             cursor.execute(qry_4, [str(batch_id), str(batch_id)])
             count = cursor.rowcount
             if count:
-                logger.info('Batch with batch_id %d set to correct recipe type revision', batch_id)
+                logger.info('Batch with batch_id %d set to correct recipe type revision (based on revision creation)',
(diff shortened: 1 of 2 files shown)


scale/scheduler/database/updater.py
```diff
@@ -143,7 +150,11 @@ def _perform_batch_field_iteration(self):
             count = cursor.rowcount
             if count:
                 logger.info('%d recipe(s) updated with batch_id %d', count, batch_id)
-            cursor.execute(qry_3, [str(batch_id), str(batch_id)])
+            cursor.execute(qry_3, [str(batch_id), str(batch_id), str(batch_id)])
+            count = cursor.rowcount
+            if count:
+                logger.info('Batch with batch_id %d set to correct recipe type revision', batch_id)
```
JohnPTobe: Should probably have a slightly different message so it's clear which condition is being triggered if we need to debug in the future.

scale/scheduler/test/database/test_updater.py
```diff
@@ -135,6 +138,22 @@ def test_update_batch_fields(self):
...
         batch_test_utils.create_batch_job(batch=batch_2, job=job_4)
 
+        # This batch tests an old batch that never created any recipes
+        time_rev_1 = now()
+        time_rev_2 = time_rev_1 + timedelta(minutes=1)
+        time_batch = time_rev_2 + timedelta(minutes=1)
+        time_rev_3 = time_batch + timedelta(minutes=1)
+        recipe_type_3 = recipe_test_utils.create_recipe_type()
+        recipe_test_utils.edit_recipe_type(recipe_type_3, recipe_type_3.definition)
+        recipe_test_utils.edit_recipe_type(recipe_type_3, recipe_type_3.definition)
```
JohnPTobe: duplicate line?

dfaller: Actually no. I need three revisions of the recipe type so I create it and then edit it twice.

PR #1154, opened 2018-04-26
Adding v6 System REST API

JohnPTobe:
Adding v6 system endpoints to REST API


scale/docs/rest/v6/system.rst
```diff
@@ -0,0 +1,324 @@
+
+.. _rest_system:
```
dfaller: Add in the v6 in the top level header

scale/docs/rest/v6/system.rst
```diff
@@ -0,0 +1,324 @@
...
+
+v6 System Services
+===============
+
+These services provide access to general system information.
+
+.. _rest_v6_system_status:
```
dfaller: Add in the sub-headings (use --- instead of ===) for each service like the v6 batch services.

scale/docs/rest/v6/system.rst
```diff
@@ -0,0 +1,324 @@
...
+
++-------------------------------------------------------------------------------------------------------------------------------+
+| **Get Scheduler Status**                                                                                                      |
++===============================================================================================================================+
+| Returns the current status of the scheduler, including information about nodes and running jobs.                              |
++-------------------------------------------------------------------------------------------------------------------------------+
+| **GET** /v6/status/                                                                                                              |
```
dfaller: Missing the right side | chars for this section. Should move the JSON responses out of the table (like in the new v6 batch API).

Issue #1093, opened 2018-01-22
Logging - Upgrade plan for Elasticsearch from 2.x to 6.x


Fizz11:
as part of refactoring the logging system, we need to create an upgrade plan to bring elastisearch, kibana, and logstash  to version 6.x

Original discussion in #1063

Fizz11:
Plan is a s follows:
1. set up a new cluster running elasticsearch 5.6 
2. restore a recent snapshot of the old 2.4 cluster
3. reindex all 2.4 data
4. point scale to the new 5.6 cluster
5. bring in data created between the first snapshot and the present, and reindex it as well
6. perform a rolling upgrade to 6.1

PR #1180, opened 2018-05-14
upgrade logstash to 5.6 and add compatability for elasticsearch 2.4-6.1

Fizz11:
issue #1093 this update provides scale with compatibility for Elasticsearch 2.4-6.1, whereas previously only 2.x was supported.


dfaller:
Hang for a bit, Mike is going to try to fix our library versions to remove the pyup issues.

dfaller:
The dependency fix went in so try merging master back into your branch to fix the pyup failure.

PR #1185, opened 2018-05-17
1144 v6 scheduler rest

JohnPTobe:
v6 REST scheduler updates


scale/docs/rest/v6/scheduler.rst
```diff
@@ -0,0 +1,100 @@
+
+.. _rest_v6_scheduler:
+
+v6 Scheduler Services
+==================
+
+These services provide access to information about the scheduler.
+There should be one and only one scheduler entry in the database and it is used to store global data like paused status.
```
dfaller: I would remove this line since it reveals system details not necessary for an API caller.

scale/docs/rest/v6/scheduler.rst
```diff
@@ -0,0 +1,100 @@
...
++----------------------+-------------------+------------------------------------------------------------------------------+
+| system_logging_level | String            | (Optional) The logging level sent to all scale system components.            |
+|                      |                   | Acceptable levels are DEBUG, INFO, WARNING, ERROR and CRITICAL.              |
+|                      |                   | Anything else will default to INFO                                           |
++--------------------------+---------------+------------------------------------------------------------------------------+
+| **Successful Response**                                                                                                 |
++--------------------+----------------------------------------------------------------------------------------------------+
+| **Status**         | 204 No content                                                                                     |
++--------------------+----------------------------------------------------------------------------------------------------+
```
dfaller: Some + characters are misaligned in this table. I know, it's a pain.

scale/scheduler/models.py
```diff
@@ -60,6 +69,12 @@ def update_scheduler(self, new_data):
         :type new_data: dict
         """
 
+        if 'resource_level' in new_data:
+            try:
+                validate_resource_level(new_data['resource_level'])
+            except ValidationError:
+                logger.exception('Invalid resource level given.')
```
dfaller: It's a good idea to include the invalid value in the exception message.

scale/scheduler/models.py
```diff
@@ -60,6 +69,12 @@ def update_scheduler(self, new_data):
...
+                raise 
         sched = self.select_for_update().filter(id=1)
         sched.update(**new_data)
```
dfaller: The `@transaction.atomic` and `select_for_update()` are unnecessary (this is very old code btw). A simple update call is sufficient (not even necessary to filter really since there is just one row).

scale/scheduler/models.py
```diff
@@ -106,6 +122,8 @@ def get_status(self):
             sched_dict['is_online'] = sched_info.is_online
             sched_dict['is_paused'] = sched.is_paused  # Note this must be pulled from the database
             sched_dict['hostname'] = sched_info.hostname
+            #TODO: Get resource level from somewhere else or set to GOOD until logic is developed
+            # sched_dict['resource_level'] = sched_info.resource_level
```
dfaller: You can set this using `sched.resource_level`. The line `sched = self.get_master()` above is retrieving the scheduler model.

scale/scheduler/test/test_views.py
```diff
@@ -22,10 +22,21 @@ def setUp(self):
         django.setup()
         Scheduler.objects.create(id=1)
 
+    def test_invalid_version(self):
+        """Tests calling the scheduler view with an invalid REST API version"""
+
+        url = '/v1/scheduler/'
+        response = self.client.generic('GET', url)
+        self.assertEqual(response.status_code, status.HTTP_404_NOT_FOUND, response.content)
+
```
dfaller: Your tests look good, just please separate the `TestSchedulerView` class into two separate test classes, one for v5 and one for v6 and split the code accordingly. This makes it easier to find, copy, and remove versions of the API.

PR #1192, opened 2018-05-21
1136 v6 node rest api changes

JohnPTobe:
Updating node rest api for v6

### DIFF
--- a/scale/docs/rest/v6/node.rst
+++ b/scale/docs/rest/v6/node.rst
@@ -63,9 +63,8 @@ Response: 200 OK
 |                    |                   |          | Duplicate it to multi-sort, (ex: order=host_name&order=created).    |
 |                    |                   |          | Prefix fields with a dash to reverse the sort, (ex: order=-created).|
 +--------------------+-------------------+----------+---------------------------------------------------------------------+
-| include_inactive   | Boolean           | Optional | If true, all nodes in the database are returned including those     |
-|                    |                   |          | marked inactive. These are typically removed from the cluster.      |
-|                    |                   |          | The default is False.                                               |
+| is_active          | Boolean           | Optional | If true, only active nodes are returned. If false, only inactive    |
+|                    |                   |          | nodes. If no value is specified, all nodes are returned.            |
 +--------------------+-------------------+----------+---------------------------------------------------------------------+
 | **Successful Response**                                                                                                 |
 +--------------------+----------------------------------------------------------------------------------------------------+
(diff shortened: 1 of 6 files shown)


scale/docs/rest/v6/node.rst
```diff
@@ -0,0 +1,239 @@
...
++--------------------+-------------------+----------+---------------------------------------------------------------------+
+| order              | String            | Optional | One or more fields to use when ordering the results.                |
+|                    |                   |          | Duplicate it to multi-sort, (ex: order=host_name&order=created).    |
+|                    |                   |          | Prefix fields with a dash to reverse the sort, (ex: order=-created).|
++--------------------+-------------------+----------+---------------------------------------------------------------------+
+| include_inactive   | Boolean           | Optional | If true, all nodes in the database are returned including those     |
+|                    |                   |          | marked inactive. These are typically removed from the cluster.      |
+|                    |                   |          | The default is False.                                               |
++--------------------+-------------------+----------+---------------------------------------------------------------------+
```
dfaller: I would recommend that we change this from `include_inactive` to `is_active` and make it a more standard filter field (no value means include all, True means only return is_active = True, etc).

scale/docs/rest/v6/node.rst
```diff
@@ -0,0 +1,239 @@
...
+       "id": 4, 
+       "hostname": "host.com", 
+       "is_paused": false, 
+       "is_active": true, 
+       "deprecated": null, 
+       "created": "2015-06-15T17:18:52.414Z", 
+       "last_modified": "2015-06-17T20:05:16.041Z", 
+       } 
+   } 
+
```
dfaller: Let's change the PATCH update to return 204 with no response body. I think this will be the standard way we should do PATCH requests.

scale/node/test/test_views.py
```diff
@@ -183,16 +221,135 @@ def test_update_active(self):
...
+        django.setup()
 
-class TestNodesStatusView(TransactionTestCase):
+        self.node1 = node_test_utils.create_node()
+        self.node2 = node_test_utils.create_node()
+        self.node3 = node_test_utils.create_node()
+
+        Scheduler.objects.create(id=1, master_hostname='localhost', master_port=5050)
+
+    @patch('mesos_api.api.get_slave')
```
dfaller: Remove all of these mock matches for v6. The v6 API should not rely on any Mesos API calls (I don't think v5 does either...)

scale/node/views.py
```diff
@@ -88,19 +121,34 @@ def get(self, request, node_id):
         return Response(serializer.data)
 
     def patch(self, request, node_id):
-        """Modify node info with a subset of fields
+        """Determine api version and call specific method
 
-        :param request: the HTTP GET request
+        :param request: the HTTP POST request
```
dfaller: `POST` should be `PATCH`

PR #1195, opened 2018-05-23
Issue 1188 - UI control for system resource utilization

### LINKED ISSUE #1188: UI control for system resource utilization
Create a UI control on the dashboard to enable users to dictate system resource utilization

ericsvendsen-mil:
Closes #1188. Added an admin-only toggle to the Overview page that lets the user switch between resource levels. Right now those are presented as "Low", "Default", and "High". Not sure what you want "Default" to be called - using "Good" seems a little vague. Maybe "Optimal"?

Anyway, here's what it looks like.

issue-1188


dfaller:
I think "Optimal" might be our best choice. We might want to include the word "Too" ("Too High") or have some additional text to avoid confusion. We wouldn't want some to hit "High" thinking they will get more resources when it really means "Too High" and resources get lowered.

PR #1229, opened 2018-06-29
1143 v6 strike

JohnPTobe:


scale/docs/rest/v6/strike.rst
```diff
@@ -0,0 +1,496 @@
...
+        "previous": null, 
+        "results": [ 
+            { 
+                "id": 1, 
+                "name": "my-strike-process", 
+                "title": "My Strike Process", 
+                "description": "This is my Strike process for detecting my favorite files!", 
+                "job": { 
+                    "id": 7, 
+                    "job_type": {
```
dfaller: Let's switch job/job type to the v6 base serializer here. Update every response affected.

scale/docs/rest/v6/strike.rst
```diff
@@ -0,0 +1,496 @@
...
+
+Response: 200 OK
+
+.. code-block:: javascript 
+ 
+    { 
+        "warnings": [ 
+            "id": "mount_change", 
+            "details": "Changing the mount path may disrupt file monitoring." 
+        ]
```
dfaller: The return for this validation should mirror batch validation. That is the standard way we'll do validation API calls.

scale/docs/rest/v6/strike.rst
```diff
@@ -0,0 +1,496 @@
...
+| description        | String            | Optional | A longer description of the Strike process.                         |
++--------------------+-------------------+----------+---------------------------------------------------------------------+
+| configuration      | JSON Object       | Optional | JSON defining the Strike configuration.                             |
+|                    |                   |          | (See :ref:`architecture_strike_spec`)                               |
++--------------------+-------------------+----------+---------------------------------------------------------------------+
+| **Successful Response**                                                                                                 |
++--------------------+----------------------------------------------------------------------------------------------------+
+| **Status**         | 200 OK                                                                                             |
++--------------------+----------------------------------------------------------------------------------------------------+
```
dfaller: Change edit API to return 204 No Content. We'll standardize on that for all "edit" API calls.

scale/docs/rest/v6/strike.rst
```diff
@@ -0,0 +1,496 @@
...
+                "id": 2 
+            }, 
+            "event": { 
+                "id": 1 
+            }, 
+            "status": "RUNNING", 
+            "priority": 10, 
+            "num_exes": 1 
+        }, 
+        "configuration": {
```
dfaller: We need a v6 Strike configuration JSON. For right now, just create a v6 JSON class (no changes from v2) and we'll worry about a separate logic class later. Do the version stripping (like in batch APIs) so the version '2.0' or '6' doesn't appear anywhere in the responses.

scale/docs/rest/v6/strike.rst
dfaller: Check out the validation warning and error classes in util.validation and see how batch uses them. Shouldn't be too big of a change.

JohnPTobe:
I'm done making the requested changes. I ended up having to create a bunch of v5 and v6 versions of ingest and job serializers to make sure the v5 stuff wasn't changed and we were getting correct versions for everything.

JohnPTobe:
@mikenholt can you do another review?

PR #1230, opened 2018-06-29
Issue 1167 - New message to process_job_input

dfaller:
Created and tested new message for process_job_input
 - will only handle a single job at a time
 - will replace current process_job_input

General Layout:
1. Query for job model with related job_type_rev, recipe and recipe__recipe_type_rev models
2. Calculate the job and recipe models that are parents of the job and query them with their output data (only if job in recipe and job does not have input data yet)
3. Combine into job's input data and validate input (if job not in recipe, input should already be there)
4. In a transaction, lock job models and bulk create job input file models
5. Create file ancestry link models
6. Update job input data and input summary fields

This message can handle both legacy and Seed job types in the recipe.
Updated message description in the Wiki.


scale/data/data/data.py
```diff
@@ -29,6 +29,24 @@ def add_value(self, value):
...
+    def add_value_from_output_data(self, input_name, output_name, output_data):
+        """Adds an output value from the given output data to this data with the given input name. This is used to pass
+        output data from a recipe node to the input data of another recipe node.
+
+        :param input_name: The name of the input value to add
+        :type input_name: string
+        :param output_name: The name of the output value in the output data
+        :type output_name: string
+        :param output_data: The output data
+        :type output_data: :class:`data.data.data.Data`
```
JohnPTobe: love this type

scale/data/data/value.py
```diff
@@ -27,6 +28,15 @@ def __init__(self, name, param_type):
         self.name = name
         self.param_type = param_type
 
+    def copy(self):
```
JohnPTobe: Not sure I understand the point of this method? Is there going to be more logic added later?

PR #1259, opened 2018-08-02
Issue 1169 - New message to create jobs

dfaller:
Created a new message for creating job(s)
Inputs:
 - Count
 - Job type name, version, and revision num
 - Event ID
 - Input data (optional, only if no recipe)
 - Root recipe ID  # Recipe containing this job
 - Recipe ID  # Recipe containing this job
 - Recipe job_name (node_name?)  # Recipe containing this job
 - Get derived superseded job from Recipe ID and Recipe job_name
 - Batch ID
 - Flag to go ahead and process input from recipe (call new message)

Updated the wiki on command messages
Fixed job type tag table name


dfaller:
You can ignore the security vulnerability, it is fixed in master.

PR #1260, opened 2018-08-02
1219 v6 patch job type

JohnPTobe:


scale/docs/rest/v6/job_type.rst
```diff
@@ -418,193 +422,195 @@ Response: 200 OK
...
+
+Request: POST http://.../v6/job-types/
+
+ .. code-block:: javascript
+     {
+      "icon_code": "f1c5",
+      "max_scheduled": 1,
+      "docker_image": "test-1.0.0-seed:1.0.0",
+      "configuration": {
+        "version": "2.0",
```
dfaller: The API should accept version 6 of the configuration. The only change here is to omit the `version` field.

scale/docs/rest/v6/job_type.rst
```diff
@@ -418,193 +422,195 @@ Response: 200 OK
...
+    { 
+		"id": 3, 
+		"name": "test", 
+		"version": "1.0.0" 
+		"title": "Job to demonstrate job type APIs", 
+		"description": "Reads input file and spit out specified number of bytes as output", 
+		"icon_code": "f1c5", 
+		"is_active": true, 
+		"is_paused": false, 
+		"is_system": true,
```
dfaller: Should flip `is_system` to `false` in doc since that should be its value.

scale/docs/rest/v6/job_type.rst
```diff
@@ -613,46 +619,160 @@ Response: 200 OK
...
 
-.. _rest_job_type_validate:
+.. _rest_v6_job_type_validate:
 
+v6 Validate Job Type
+--------------------
+
+**Example POST /v6/job-types/ API call**
+
+Request: POST http://.../v6/job-types/
```
dfaller: The `/validation/` is missing from the end of the example URLs here.

scale/docs/rest/v6/job_type.rst
```diff
@@ -840,579 +792,80 @@ Response: 200 OK
...
+Request: PATCH http://.../v6/job-types/test/1.0.0/
+
+ .. code-block:: javascript  
+    {
+      "icon_code": "012F",
+      "is_active": true,
+      "is_paused": false,
+      "max_scheduled": 1,
+      "configuration": {
+        "version": "2.0",
```
dfaller: Same comment as above about removing `version` from the configuration JSON. This should be done everywhere in the v6 Job Type docs.

scale/job/models.py
```diff
@@ -2460,13 +2465,16 @@ def create_job_type_v6(self, manifest_dict, trigger_rule_dict=None, configuratio
...
         self._validate_job_type_fields(**kwargs)
 
-        trigger_rule = self._create_seed_job_trigger_rule(manifest, trigger_rule_dict)
+        trigger_rule = None
+        if trigger_rule_dict:
+            trigger_rule = self._create_seed_job_trigger_rule(manifest, trigger_rule_dict)
 
         secrets = None
         if configuration_dict:
             configuration = JobConfigurationV6(configuration_dict, do_validate=True).get_configuration()
```
dfaller: I recommend moving this line, `configuration = JobConfigurationV6(configuration_dict, do_validate=True).get_configuration()` into the view (within a try/except for `InvalidConfiguration`) and pass the general configuration object into this method instead of passing the dict.

scale/job/models.py
```diff
@@ -2460,13 +2465,16 @@ def create_job_type_v6(self, manifest_dict, trigger_rule_dict=None, configuratio
                 raise Exception('%s is not an editable field' % field_name)
         self._validate_job_type_fields(**kwargs)
 
-        trigger_rule = self._create_seed_job_trigger_rule(manifest, trigger_rule_dict)
+        trigger_rule = None
+        if trigger_rule_dict:
+            trigger_rule = self._create_seed_job_trigger_rule(manifest, trigger_rule_dict)
```
dfaller: Go ahead and remove trigger_rule_dict from this method. v6 will not support trigger rules.

dfaller: Also remove `**kwargs`. We should only allow specific parameters into this method and set specific fields on the job type model.

dfaller: Also remove `_create_seed_job_trigger_rule()`

dfaller: Also make the same changes to `edit_job_type_v6()`. For editing, remove trigger rule and kwargs, and make sure editable fields like is_active are passed in. Also pass in the configuration object just like `create_job_type_v6()`. Pass in the manifest class as well instead of the dict and make sure the manifest updates fields like title, description etc when editing.

scale/job/models.py
```diff
@@ -3184,6 +3193,73 @@ def validate_job_type_v5(self, name, version, interface, error_mapping=None, tri
...
+                warnings.extend(config.validate(manifest))
+            except InvalidJobConfiguration as ex:
+                is_valid = False
+                errors.append(ex.error)
+                message = 'Job type configuration invalid'
+                logger.exception(message)
+                pass
+
+        if manifest:
+            try:
```
dfaller: This validation block checking against recipe types can be removed. My new recipe work now pins recipes to a job type revision, so changing a job type never breaks a recipe now.

scale/job/models.py
dfaller: How does `create_job_type_v6()` work with existing job types? It seems like it always saves a new one which would error when an existing job type with the same name and version already exists.

EDIT: Never mind, I see you split the logic in the view. That's fine.

scale/job/views.py
```diff
@@ -277,57 +278,75 @@ def create_v6(self, request):
...
-            message = 'Job type custom resources invalid'
-            logger.exception(message)
-            raise BadParameter('%s: %s' % (message, unicode(ex)))
+        name = manifest_dict['job']['name']
+        version = manifest_dict['job']['jobVersion']
+        
+        existing_job_type = JobType.objects.filter(name=name, version=version).first()
+        if not existing_job_type:
+            try:
+                with transaction.atomic():
```
dfaller: The `with transaction.atomic():` is unneeded here.

scale/job/views.py
```diff
@@ -277,57 +278,75 @@ def create_v6(self, request):
...
+                raise BadParameter('%s: %s' % (message, unicode(ex)))
+            except InvalidSeedManifestDefinition as ex:
+                message = 'Job type interface invalid'
+                logger.exception(message)
+                raise BadParameter('%s: %s' % (message, unicode(ex)))
+            except InvalidJobConfiguration as ex:
+                message = 'Job type configuration invalid'
+                logger.exception(message)
+                raise BadParameter('%s: %s' % (message, unicode(ex)))
+            except InvalidResources as ex:
```
dfaller: Should be able to remove `InvalidResources` except.

scale/job/views.py
```diff
@@ -277,57 +278,75 @@ def create_v6(self, request):
...
+            try:
+                with transaction.atomic():
+    
+                    # Create the job type
+                    job_type = JobType.objects.create_job_type_v6(manifest_dict=manifest_dict,
+                                                                  trigger_rule_dict=None,
+                                                                  configuration_dict=configuration_dict,
+                                                                  **extra_fields)
+    
+            except (InvalidJobField, InvalidTriggerType, InvalidTriggerRule, InvalidTriggerMissingConfiguration,
```
dfaller: Remove the `**extra_fields` and remove catching these exceptions, except `InvalidSecretsConfiguration`.

scale/job/views.py
```diff
@@ -277,57 +278,75 @@ def create_v6(self, request):
...
+            except InvalidResources as ex:
+                message = 'Job type custom resources invalid'
+                logger.exception(message)
+                raise BadParameter('%s: %s' % (message, unicode(ex)))
+        else:
+            try:
+                with transaction.atomic():
+                    # Edit the job type
+                    JobType.objects.edit_job_type_v6(job_type_id=existing_job_type.id, manifest_dict=manifest_dict,
+                                                     configuration_dict=configuration_dict, **extra_fields)
```
dfaller: Same changes as comment above for creating a job type.

scale/docs/rest/v6/job_type.rst
JohnPTobe: Should it be removed or just changed to 6 as these are POST examples?

JohnPTobe:
I addressed all the issues from the PR but moving the configuration logic as I feel the logic belongs inside the model and it saves passing the secrets as well as the configuration.

scale/docs/rest/v6/job_type.rst
dfaller: Just remove the `"version": "2.0",` line, since the rest is still version 6 compliant.

scale/job/models.py
```diff
@@ -2435,53 +2435,45 @@ def create_job_type_v5(self, name, version, interface, trigger_rule=None, error_
...
         # Create the new job type
         job_type = JobType()
-        
-        secrets = None
-        if configuration_dict:
-            configuration = JobConfigurationV6(configuration_dict, do_validate=True).get_configuration()
-            configuration.validate(manifest)
-            secrets = configuration.remove_secret_settings(manifest)
-            configuration_dict = convert_config_to_v6_json(configuration).get_dict()
-            job_type.configuration = configuration_dict
```
dfaller: Sorry if I wasn't clear. My intent was for the view to have just the line converting the v6 JSON:
`configuration = JobConfigurationV6(configuration_dict, do_validate=True).get_configuration()`

This method would then contain the rest of the code, something like this:
```
if not configuration:
    configuration = JobConfiguration()
configuration.validate(manifest)
secrets = configuration.remove_secret_settings(manifest)
job_type.configuration = convert_config_to_v6_json(configuration).get_dict()

```

This way the view is handling the JSON validation logic with dealing with user input and the manager method is handling business logic/validation and operates on the logic class (`JobConfiguration`).

scale/job/models.py
```diff
@@ -2643,6 +2648,11 @@ def edit_job_type_v6(self, job_type_id, manifest=None, docker_image=None, icon_c
             job_type.author_url = manifest.get_maintainer()['url']
             job_type.save()
```
dfaller: After the job type save here, we need a call to `JobTypeRevision.objects.create_job_type_revision(job_type)` just like during job type creation so there is a new revision for the manifest changes. The call must be after the job type save because the call assumes that `manifest` and `revision_num` have been updated.

dfaller: Never mind, it's there. :)

PR #1261, opened 2018-08-06
1134 v6 error

JohnPTobe:
Keep cranking out those v6 updates!


scale/docs/rest/v6/error.rst
```diff
@@ -0,0 +1,153 @@
...
+        "results": [ 
+            { 
+                "id": 1, 
+                "name": "unknown", 
+                "title": "Unknown", 
+                "description": "The error that caused the failure is unknown.", 
+                "category": "SYSTEM", 
+                "is_builtin": true, 
+                "created": "2015-03-11T00:00:00Z", 
+                "last_modified": "2015-03-11T00:00:00Z"
```
dfaller: Let's go ahead and include `job_type_name` as a field for both responses.

PR #1266, opened 2018-08-08
1142 v6 scan

JohnPTobe:
Implementation of v6 scan api


scale/docs/rest/v6/scan.rst
```diff
@@ -0,0 +1,672 @@
...
+|                    |                   | (See :ref:`Job Details <rest_v6_job_details>`)                                 |
++--------------------+-------------------+--------------------------------------------------------------------------------+
+| created            | ISO-8601 Datetime | When the associated database model was initially created.                      |
++--------------------+-------------------+--------------------------------------------------------------------------------+
+| last_modified      | ISO-8601 Datetime | When the associated database model was last saved.                             |
++--------------------+-------------------+--------------------------------------------------------------------------------+
+| configuration      | JSON Object       | JSON defining the Scan configuration.                                          |
+|                    |                   | (See :ref:`architecture_scan_spec`)                                            |
++--------------------+-------------------+--------------------------------------------------------------------------------+
```
dfaller: At the end of the docs here, we need to include the v6 JSON schema for the scan configuration. All references to the old configuration (`:ref:architecture_scan_spec`) need to be updated to the v6 version here.

We need to have a base logic class for the ScanConfiguration. I see you already made a v6 JSON schema class for the scan configuration. The validate() logic should get moved into the logic class. The JSON schema classes (v1 and v6) can have a method to return the logic class. The logic class should get passed around to handle all business. Add methods to the Scan model to handle getting the ScanConfiguration logic class and JSON dicts for v1 and v6. See recipe/definition and the recipe type model for an example.

I just realized I missed this on the v6 Strike API as well.

JohnPTobe:
here are the commits since the last review:
https://github.com/ngageoint/scale/compare/a610d16d56b6189d14937e5c4abc338e41ae3179...1142-v6Scan

JohnPTobe:
I made the logic/json changes for strike config btw

Issue #1207, opened 2018-06-07
FIle Purge - Remove records related to the Job model


mikenholt:
Another step in the file purge process will be scrubbing Jobs associated with files (`F_ids`) that we need to purge. This will involve:
  - Remove `JobInputFile` and `FileAncestryLink` records related to the `F_ids`
    - If any of the `F_ids` are an ancestor in the `FileAncestryLink` table then all descendants will have to have to be added to the `F_ids` list.
  - Remove `JobExecutionOutput`, `JobExecutionEnd` and `JobExecution` records related to the `Job` records that will have to be removed. 
  - Remove `Job` records related to the `JobInputFile` records that are removed.

dfaller:
I think the next step is a purge_jobs message (take up to 100 job IDs).
Purge the `FileAncestryLink` (just match on job field), `TaskUpdate`, `JobExecutionOutput`, `JobExecutionEnd`, `JobExecution`, `BatchJob`, `RecipeNode`, `JobInputFile` and `Queue` models that match the jobs to be purged, then purge the job models.

This new message should be kicked off from a delete_files message (when purging). If purging, grab the job IDs of the files to purge and send a purge_jobs message for them (need to consider how to handle message failure).

PR #1281, opened 2018-08-16
#1207 - File Purge - Remove records related to the Job model 

mikenholt:
- Created a new message for the Job app that will purge job (and related) models.  
- Models that are deleted: `JobExecution`, `TaskUpdate`, `JobExecutionOutput`, `JobExecutionEnd`, `FileAncestryLink`, `BatchJob`, `RecipeNode`, `JobInputFile`, `Queue`, and `Job`
- The creation of this message is called from the `delete_files` message if a system purge is requested. 

I updated the command-messages wiki

### DIFF
--- a/scale/storage/messages/delete_files.py
+++ b/scale/storage/messages/delete_files.py
@@ -102,7 +102,7 @@ class DeleteFiles(CommandMessage):
 
         if self.purge:
             files_to_delete.delete()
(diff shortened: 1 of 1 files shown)


dfaller:
I think merging master into your branch will fix the vulnerability problem.

PR #1309, opened 2018-08-28
Issue 1307 - Fixed submitting Seed manifest without tags

dfaller:
I fixed the bug so that minimum Seed manifests are correctly accepted by the v6 REST API.

If approved, please immediately merge and delete branch.


mikenholt:
Fixes #1309

PR #1312, opened 2018-08-30
Support for pulling docker_image from JobTypeRevision

gisjedi:
Removed logic to compute docker_image at runtime in favor of storing each update on job_type_rev.


scale/job/execution/configuration/configurators.py
```diff
@@ -386,7 +386,7 @@ def _configure_regular_job(config, job_exe, job_type):
         """
 
         config.create_tasks(['pull', 'pre', 'main', 'post'])
-        config.add_to_task('pull', args=create_pull_command(job_type.get_tagged_docker_image()))
+        config.add_to_task('pull', args=create_pull_command(job_exe.job.job_type_rev.docker_image))
```
dfaller: This will result in 2 database queries for every job_exe scheduled. Recommend adding `docker_image` field to the queue model so it is available for scheduling without additional queries.

scale/job/migrations/0045_auto_20180830_1812.py
```diff
@@ -0,0 +1,30 @@
...
+
+    dependencies = [
+        ('job', '0044_jobtypetag_job_type'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='jobexecution',
+            name='docker_image',
+            field=models.TextField(default=''),
```
dfaller: This field should be nullable, otherwise it requires a rewrite of the entire table.

scale/job/migrations/0045_auto_20180830_1812.py
```diff
@@ -0,0 +1,30 @@
...
+        ),
+        migrations.AddField(
+            model_name='jobtyperevision',
+            name='docker_image',
+            field=models.TextField(default='not-a-real-image'),
+        ),
+        migrations.AlterField(
+            model_name='jobtype',
+            name='docker_image',
+            field=models.CharField(default='', max_length=500),
```
dfaller: What was this field altered for?

scale/queue/job_exe.py
```diff
@@ -45,6 +45,7 @@ def create_job_exe_model(self, framework_id, when):
         job_exe.batch_id = self._queue.batch_id
         job_exe.exe_num = self._queue.exe_num
         job_exe.timeout = self._queue.timeout
+        job_exe.docker_image = self._queue.docker_image
```
dfaller: This either needs to have a fall back or we need to populate the `docker_image` field in the `queue` table with a migration prior to execution. Otherwise existing jobs on the queue will fail with blank Docker image names,

scale/job/migrations/0045_auto_20180830_1812.py
gisjedi: Without a default, all unit tests fail as its a nullable field and jobtyperevision is not.

scale/job/models.py
```diff
@@ -1933,6 +1933,7 @@ class JobExecution(models.Model):
     exe_num = models.IntegerField(blank=True, null=True)
     cluster_id = models.CharField(blank=True, max_length=100, null=True)
     node = models.ForeignKey('node.Node', blank=True, null=True, on_delete=models.PROTECT)
+    docker_image = models.TextField(default='')
```
dfaller: This still needs to be nullable with no default.

Issue #1284, opened 2018-08-21
Remove resource level from the v6 scheduler REST APIs


dfaller:
Remove resource level from the v6 scheduler REST APIs. Also remove the resource_level field from the scheduler model.

Issue #1221, opened 2018-06-14
Other v6 Job-related REST APIs


dfaller:
**Job with Execution List**
GET /jobs/executions/ - this should be totally removed in v6

**Retrieve Job Input File List**
GET /jobs/{id}/input_files/
Query params can stay the same as v5, change return to match v6 /files/ response (docs can reference v6 /files/ response instead of duplicating).

**Retrieve Job Executions List** (replace what is currently there)
GET /jobs/{id}/executions/
Query Params
- status (allow multiple)
- node_id (allow multiple)
- error_id (allow multiple)
- error_category (allow multiple)

Response 200 OK, use paged list, use v6 job execution serializer (re-use current `JobExecutionSerializer`), order by exe_num desc

**Retrieve Job Execution Details** (replace what is currently there)
GET /jobs/{id}/executions/{exe_num}/
Response 200 OK, use v6 job execution details serializer (re-use current `JobExecutionDetailsSerializer`)

**Cancel Jobs**
POST /jobs/cancel/
Query Params - include current v6 query params, add `is_superseded`, `batch_id`, `recipe_id`, and `job_type_name` to match v6 /jobs/
Response 202 Accepted

**Requeue Jobs**
POST /jobs/requeue/
Query Params - include current v6 query params, add `is_superseded`, `batch_id` `recipe_id`, and `job_type_name` to match v6 /jobs/
Response 202 Accepted

These logging APIs (put in job REST APIs) should be carried over to v6 (same as v5):
GET /job-executions/{id}/logs/stdout/
GET /job-executions/{id}/logs/stderr/
GET /job-executions/{id}/logs/combined/

All other /job-executions/...  APIs should be removed in v6.

PR #1315, opened 2018-08-31
1221 other v6 job

JohnPTobe:
Addresses #1221 and #1284


JohnPTobe:
Ok, I held off on the logging docs as there was nothing for v5 and I couldn’t find a sample response.

Issue #1209, opened 2018-06-07
Delete job files message


dfaller:
Create a message that takes a job ID and a purge flag. This message will look at the job's files and create and queue a system job to delete those files in the workspace (passing along the purge flag). This messages does not affect any database models.

PR #1324, opened 2018-09-13
#1209 - Delete job files message

mikenholt:
- Added JobTypeRev fixture for `scale-delete-files`
- Added command message `spawn-delete-files-job` for creating and queuing a system job that will delete files in workspaces

I will update the wiki after approval but before merge.

### DIFF
--- a/scale/job/messages/spawn_delete_files_job.py
+++ b/scale/job/messages/spawn_delete_files_job.py
@@ -1,6 +1,7 @@
 """Defines a command message that creates and queues a system job for deleting files"""
 from __future__ import unicode_literals
 
+import json
 import logging
 
 from data.data.data import Data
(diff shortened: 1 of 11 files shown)


scale/job/messages/spawn_delete_files_job.py
```diff
@@ -0,0 +1,102 @@
...
+        return {'job_id': self.job_id, 'trigger_id': self.trigger_id, 'purge': str(self.purge)}
+
+    @staticmethod
+    def from_json(json_dict):
+        """See :meth:`messaging.messages.message.CommandMessage.from_json`
+        """
+
+        message = SpawnDeleteFilesJob()
+        message.job_id = json_dict['job_id']
+        message.trigger_id = json_dict['job_id']
```
JohnPTobe: copypasta error?

scale/job/seed/manifest.py
```diff
@@ -207,7 +207,7 @@ def get_input_interface(self):
         """
 
         input_dict = self.get_inputs()
```
JohnPTobe: remove unused input_dict?

mikenholt: I made this change at the request of Derick.  @dfaller any thoughts?

Issue #1247, opened 2018-07-26
Fix Seed missing output error


dfaller:
Right now if a Seed job does not provide an expected (required) output that matches an output file pattern, it just throws an exception and causes an UNKNOWN system error. Fix this so that a missing required output from a Seed job gets registered as a `missing-required-output` error.

PR #1331, opened 2018-09-17
Issue 1247

emimaesmith:
#1247 
Replaced the ScaleOutputError with a MissingRequiredOutput error when a required output file is not found. 
Added a new UnexpectedMultipleOutputs error that is raised when multiple files are found for a singular output.


scale/scale/settings.py
```diff
@@ -80,7 +80,7 @@
 # SECURITY WARNING: don't run with debug turned on in production!
 DEBUG = False
 
-ALLOWED_HOSTS = ['localhost', '127.0.0.1']
+ALLOWED_HOSTS = ['localhost', '127.0.0.1', 'scale-dev-emimaesmith.c9users.io']
```
JohnPTobe: Probably should change this back.

JohnPTobe: You can override it in your local settings file

emimaesmith:
@JohnPTobe Good catch. Fixed the settings file.

Issue #1210, opened 2018-06-07
Recipe purge


dfaller:
Create a purge_recipe message that takes a single recipe ID:
 - message should retrieve the recipe instance and determine the current, original leaf nodes (is_original == True), send delete_job_files messages (with purge) for original leaf jobs and send purge_recipe messages for original leaf recipes
 - if there are no original nodes left in the recipe, delete `BatchRecipe`, `RecipeNode` (matched on sub_recipe or recipe), `RecipeInputFile`, and `Recipe` model for this recipe ID
 - if this recipe had a higher-level recipe containing it (had an is_original == True `RecipeNode` model matched on sub_recipe), send a purge_recipe message for the higher-level recipe
 - if the recipe did not have a higher-level recipe (is top-level) but has superseded another recipe, send a purge_recipe message for the superseded recipe

Update the purge_jobs message to send a purge_recipe message if the job had an is_original == True `RecipeNode` model attached to it. This way when a job is purged, it will alert its original recipe which can then purge that job's parents.

mikenholt:
fixed with #1339

PR #1339, opened 2018-09-24
#1210 - Recipe purge

mikenholt:
- Added `get_original_leaf_nodes` to `RecipeInstance`
- Added `trigger_id` passing to the rest of the purge stack
- Modified 'purge_jobs' to so that if the job had an is_original == True `RecipeNode` model attached to it it will pass a message for `purge_recipe`
- Added logic for purging recipes and their contents.  Expected behavior is: 
  - message should retrieve the recipe instance and determine the current, original leaf nodes (is_original == True), send delete_job_files messages (with purge) for original leaf jobs and send purge_recipe messages for original leaf recipes
  - if there are no original nodes left in the recipe, delete `BatchRecipe`, `RecipeNode` (matched on sub_recipe or recipe), `RecipeInputFile`, and `Recipe` model for this recipe ID
  - if this recipe had a higher-level recipe containing it (had an is_original == True `RecipeNode` model matched on sub_recipe), send a purge_recipe message for the higher-level recipe
  - if the recipe did not have a higher-level recipe (is top-level) but has superseded another recipe, send a purge_recipe message for the superseded recipe

**I will update the command messages wiki after approval but before merge**

### DIFF
--- a/scale/recipe/instance/recipe.py
+++ b/scale/recipe/instance/recipe.py
@@ -61,7 +61,7 @@ class RecipeInstance(object):
         leaf_job_ids = []
         leaf_recipe_ids = []
 
-        for node in self.graph:
+        for _, node in self.graph.items():
             if node.is_original:
                 if node.node_type == JobNodeDefinition.NODE_TYPE:
                     leaf_job_ids.append(node.job.id)
--- a/scale/recipe/instance/recipe.py
+++ b/scale/recipe/instance/recipe.py
@@ -58,14 +58,6 @@ class RecipeInstance(object):
         :rtype: dict
         """
 
-        leaf_job_ids = []
-        leaf_recipe_ids = []
+        leaf_nodes = {n.name: n for n in self.graph.values() if n.is_original and not n.children}  # {Name: Node}
 
-        for _, node in self.graph.items():
-            if node.is_original:
-                if node.node_type == JobNodeDefinition.NODE_TYPE:
-                    leaf_job_ids.append(node.job.id)
-                elif node.node_type == RecipeNodeDefinition.NODE_TYPE:
-                    leaf_recipe_ids.append(node.recipe.id)
-
-        return {'jobs': leaf_job_ids, 'recipes': leaf_recipe_ids}
+        return leaf_nodes
--- a/scale/recipe/instance/recipe.py
+++ b/scale/recipe/instance/recipe.py
@@ -51,5 +51,21 @@ class RecipeInstance(object):
 
         return {'BLOCKED': blocked_job_ids, 'PENDING': pending_job_ids}
 
-    # Function to find leaf recipes and jobs 
-    # returns leaf nodes 
+    def get_original_leaf_nodes(self):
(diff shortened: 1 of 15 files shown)


scale/recipe/instance/recipe.py
```diff
@@ -50,3 +50,22 @@ def get_jobs_to_update(self):
...
+        leaf_recipe_ids = []
+
+        for _, node in self.graph.items():
+            if node.is_original:
+                if node.node_type == JobNodeDefinition.NODE_TYPE:
+                    leaf_job_ids.append(node.job.id)
+                elif node.node_type == RecipeNodeDefinition.NODE_TYPE:
+                    leaf_recipe_ids.append(node.recipe.id)
+
+        return {'jobs': leaf_job_ids, 'recipes': leaf_recipe_ids}
```
dfaller: I would recommend returning a dict where the original leaf node names map to the instance node. The function will be more general and reusable that way. Also you need a check in here for the node's children to be empty, otherwise it is not limited to leaf nodes. Also if you are just iterating over the node instances (and not using node name) you can use `self.graph.values()` instead of `self.graph.items()`.

scale/recipe/messages/purge_recipe.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+def create_purge_recipe_message(recipe_id, trigger_id):
+    """Creates messages to remove the given recipe by ID
+
+    :param recipe_id: The recipe ID
+    :type purge_job_ids: int
+    :param trigger_id: The trigger event id for the purge operation
+    :type trigger_id: int
+    :return: The list of messages
```
dfaller: Minor comment typo. Also the description says `messages`.

scale/recipe/messages/purge_recipe.py
```diff
@@ -0,0 +1,101 @@
...
+                RecipeNode.objects.filter(sub_recipe=recipe).delete()
+
+        # Kick off purge_recipe for a superseded recipe
+        elif recipe.superseded_recipe:
+            self.new_messages.append(create_purge_recipe_message(recipe_id=recipe.superseded_recipe.id,
+                                                                 trigger_id=self.trigger_id))
+
+        # Delete BatchRecipe, RecipeNode, RecipeInputFile, and Recipe
+        BatchRecipe.objects.filter(recipe=recipe).delete()
+        RecipeNode.objects.filter(recipe=recipe).delete()
```
dfaller: Need to match on sub_recipe too: `RecipeNode.objects.filter(Q(recipe=recipe) | Q(sub_recipe=recipe)).delete()`

scale/recipe/messages/purge_recipe.py
```diff
@@ -0,0 +1,101 @@
...
+        RecipeInputFile.objects.filter(recipe=recipe).delete()
+        recipe.delete()
```
dfaller: Lines 83 - 99 should only be executed if there are no original leaf nodes. Basically you don't want to delete this recipe (and start purge of its parent/superseded recipe) until all of its original leaf nodes are gone.

Issue #1211, opened 2018-06-07
Purge source file message


dfaller:
Create a purge_source_file message that takes a single source file ID. The message should:
1. If there are any non-superseded recipes with this file as input, send purge_recipe messages for them
2. If there are any non-recipe jobs with this file as input, send delete_job_files messages for them
3. If there are zero from (1) and (2), delete the `Ingest` and `ScaleFile` models for the source file. This is a fully successful purge!

Update the purge_jobs and purge_recipe messages:
1. The purge_jobs message should send purge_source_file messages for all source file inputs going into non-recipe jobs.
2. The purge_recipe message should send purge_source_file messages for all source file inputs going into it if it is its own root superseded recipe (root_superseded_recipe is null or equal to ID).

PR #1341, opened 2018-09-26
#1211 - Purge source file message

mikenholt:
- Added `purge_source_file` message type with expected `execute()` behavior of: 
  ```
  1. If there are any non-superseded recipes with this file as input, 
       send purge_recipe messages for them
  2. If there are any non-recipe jobs with this file as input, 
       send delete_job_files messages for them
  3. If there are zero from (1) and (2), 
       delete the `Ingest` and `ScaleFile` models for the source file. 
  ```
- Added `purge` flag passing to `purge_recipe`, `purge_jobs` and `delete_files`
- Modified `purge_job` to send a `purge_source_file` message for any input files if the job is not part of a recipe.
- Modified `purge_recipe` to send a `purge_source_file` message for any input files if it is its own root superseded recipe

**I will update the command messages wiki page after review but before merge**

### DIFF
--- a/scale/job/messages/purge_jobs.py
+++ b/scale/job/messages/purge_jobs.py
@@ -114,7 +114,7 @@ class PurgeJobs(CommandMessage):
 
         # Kick off purge_source_file for all source file inputs
         input_source_files = JobInputFile.objects.filter(job__in=self._purge_job_ids, job__recipe__isnull=True)
-        input_source_files.select_relate('input_file')
+        input_source_files.select_related('input_file')
         for source_file in input_source_files:
             self.new_messages.append(create_purge_source_file_message(source_file_id=source_file.input_file.id,
                                                                       trigger_id=self.trigger_id,
--- a/scale/job/messages/purge_jobs.py
+++ b/scale/job/messages/purge_jobs.py
@@ -24,15 +24,13 @@ MAX_NUM = 100
 logger = logging.getLogger(__name__)
 
 
-def create_purge_jobs_messages(purge_job_ids, trigger_id, purge):
+def create_purge_jobs_messages(purge_job_ids, trigger_id):
     """Creates messages to remove the given job IDs
 
     :param purge_job_ids: The job IDs
     :type purge_job_ids: list
     :param trigger_id: The trigger event id for the purge operation
     :type trigger_id: int
-    :param purge: Boolean value to determine if files should be purged from workspace
-    :type purge: bool
     :return: The list of messages
     :rtype: list
     """
@@ -48,7 +46,6 @@ def create_purge_jobs_messages(purge_job_ids, trigger_id, purge):
             message = PurgeJobs()
         message.add_job(job_id)
         message.trigger_id = trigger_id
-        message.purge = purge
     if message:
         messages.append(message)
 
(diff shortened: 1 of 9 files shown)


scale/job/messages/purge_jobs.py
```diff
@@ -106,11 +112,20 @@ def execute(self):
         """See :meth:`messaging.messages.message.CommandMessage.execute`
         """
 
-        # Kick off a purge_recipe for recipe with node job
+        # Kick off purge_source_file for all source file inputs
+        input_source_files = JobInputFile.objects.filter(job__in=self._purge_job_ids, job__recipe__isnull=True)
+        input_source_files.select_related('input_file')
```
dfaller: Need to filter to make sure you only grab source files.

scale/job/messages/purge_jobs.py
```diff
@@ -64,6 +68,7 @@ def __init__(self):
         self._count = 0
         self._purge_job_ids = []
         self.trigger_id = None
+        self.purge = False
```
dfaller: Do you need a purge flag for this message? Since it is purging, you know purge will always be true.

scale/recipe/messages/purge_recipe.py
```diff
@@ -7,28 +7,32 @@
...
+    :param purge: Boolean value to determine if files should be purged from workspace
+    :type purge: bool
     :return: The purge recipe message
     :rtype: :class:`recipe.messages.purge_recipe.PurgeRecipe`
     """
 
     message = PurgeRecipe()
     message.recipe_id = recipe_id
     message.trigger_id = trigger_id
+    message.purge = purge
```
dfaller: Do you need a purge flag for this message? Since it is purging, you know purge will always be true.

scale/recipe/messages/purge_recipe.py
```diff
@@ -68,6 +74,16 @@ def execute(self):
         """
 
         recipe = Recipe.objects.select_related('superseded_recipe').get(id=self.recipe_id)
+
+        # Kick off purge_source_file for source file inputs of the given recipe
+        input_source_files = RecipeInputFile.objects.filter(Q(recipe__root_superseded_recipe=recipe) |
+                                                            Q(recipe__root_superseded_recipe__isnull=True))
+        input_source_files.select_related('input_file')
```
dfaller: I just realized that this (and job logic as well) only works if the jobs and recipes involved are run with a single source file. If multiple source files kick off these jobs/recipes, we will end up purging all of them. We should a note (new issue?) to address this in the future.

scale/source/messages/purge_source_file.py
```diff
@@ -0,0 +1,98 @@
...
+
+    def __init__(self):
+        """Constructor
+        """
+
+        super(PurgeSourceFile, self).__init__('purge_source_file')
+
+        self.source_file_id = None
+        self.trigger_id = None
+        self.purge = False
```
dfaller: Again I don't think you need the purge flag since the point of the message is to do a purge.

scale/recipe/messages/purge_recipe.py
mikenholt: I opened #1342.  Feel free to weigh in with your thoughts 💯 👌

Issue #1177, opened 2018-05-10
Use new recipe logic


dfaller:
This issue will require moving towards unique recipe type names. Add migrations to fix currently conflicting recipe type names and add a unique index to recipe_type.name. Also update the recipe type creation logic (both API and test utils) to correctly deconflict recipe type names. - DONE

Update to message update_recipe_metrics: - DONE
 - send update_recipe message to higher-level (containing) recipe if applicable

Switch out all uses of update_recipes to use update_recipe message instead - DONE

Switch out all uses of reprocess_recipes (all additional recipe logic, will include updating batches and batch definition with recursive node_name force reprocess) to use create_recipes - DONE

When updating job status messages to use new update_recipe message, make sure to send a "force all" forced nodes through. - DONE

When updating process_recipe_input message to use update_recipe message, make sure to test passing along forced_nodes. - DONE

Complete major unit test to create_recipes that goes through a major re-process containing sub-recipes - DONE

Make sure batches correctly only reprocess the top level recipes (not sub-recipes) from a previous batch. - DONE
Update batch docs to include new forced_nodes. - DONE

Update recipe type models to save/retrieve recipe definitions in correct JSON versions. - DONE

Search through all recipe model and handler code and remove/deprecate all old stuff - DONE

Try to eliminate use of get_revision_old() in the JobTypeManager and RecipeTypeManager - DONE

After this issue is complete, design out the details for the issue for the v6 recipe type and recipe REST APIs. When creating a new job/recipe revision, the job type/recipe type APIs should have a flag that automatically pushes latest revision to containing recipes recursively. Will need a command message to handle this.

PR #1346, opened 2018-10-01
Issue 1171 - New message to update recipes

dfaller:
- Updated `create_jobs` message so all of a recipe's jobs could be created in one message
 - Created new `update_recipe` message which handles updating a single recipe and supports sub-recipes
 - Updated `process_recipe_input` and `create_recipes` messages to pass along forced node information to the resulting `update_recipe` message


scale/recipe/messages/process_recipe_input.py
```diff
@@ -84,6 +97,7 @@ def execute(self):
             Recipe.objects.process_recipe_input(recipe)
 
         # Create message to update the recipe
+        # TODO: switch to new update_recipe message and send forced_nodes
```
mikenholt: Did you mean to implement this change or is this for a future issue?

dfaller: Future issue: #1177

mikenholt: 👌

Issue #1126, opened 2018-03-14
Provide purge results


mikenholt:
Once a file purge has been completed a summary of what was removed (source files, products, logs, etc.) should be made available for reporting purposes.

mikenholt:
Notes: Add a `PurgeResults` model to the `storage` app of Scale.  

Potential columns: 
  - `source_file_id`
  - `num_jobs_deleted`
  - `num_recipes_deleted`
  - `num_products_deleted`
  - `purge_started`
  - `purge_completed`
  - `user` foreign key (from REST API `purge-source` call)

Add to the relevant fields as the messages are being processed.

PR #1356, opened 2018-10-04
#1126 - Provide purge results

mikenholt:
- Added `PurgeResults` model to the `storage` app
- Added relevant updaters to the various messages so that the purge process is tracked properly. 

**I will update the command messages wiki after acceptance but before merge**

💥 💥 💥 💥 💥 💥 💥 💥 💥 💥 💥 💥 💥 💥 💥


mikenholt:
The test build failed with a timeout so I will rerun 🚀

Issue #1395, opened 2018-10-30
Add ability to stop purge


mikenholt:
Add a field to purge_results that will be checked at the beginning of each message associated with a purge. If a certain condition (`true`|`false`), `return true` from the `execute()`.

PR #1401, opened 2018-11-01
#1395 - Add ability to stop purge

mikenholt:
- Added `force_stop_purge` to `PurgeResults`
- Revised how `PurgeResults` was queried and set if something needed to be updated.
- Added tests


mikenholt:
My thinking was the `force_stop_purge` field could just be set manually in the DB if a purge operation needs to be purged.  That seems less than ideal for some situations so if we want that to be set via another POST request, we can open an issue for that.

scale/job/test/messages/test_purge_jobs.py
```diff
@@ -39,29 +43,12 @@ def test_json(self):
...
-        message.status_change = timezone.now()
-
-        # Execute message
-        result = message.execute()
-        self.assertTrue(result)
-
-        # Check that job is deleted
-        self.assertEqual(Job.objects.filter(id=job.id).count(), 0)
-        self.assertEqual(JobExecution.objects.filter(id=job_exe.id).count(), 0)
-
```
JohnPTobe: Is this test redundant now?

scale/job/test/messages/test_purge_jobs.py
```diff
@@ -137,3 +126,34 @@ def test_execute_check_results(self):
         # Check results are accurate
         self.assertEqual(PurgeResults.objects.values_list('num_jobs_deleted', flat=True).get(
             source_file_id=source_file.id), 3)
+
+    def test_execute_force_stop_purge(self):
+        """Tests calling PurgeJobs.execute() successfully"""
```
JohnPTobe: Update comments for tests

Issue #1383, opened 2018-10-18
Mesos Roles


nevasway:
It would be helpful if Scale supported Mesos Roles for managing resource allocation.

gisjedi:
This issue has ongoing work needed to fully support the capability found under #1459

Issue #1040, opened 2017-12-18
Update scheduler to use Mesos HTTP v1 Scheduler API


dfaller:
Update the scheduler to use the Mesos HTTP v1 Scheduler API.

gisjedi:
This is being worked as a requirement of supporting DCOS Strict mode for some clients. Supporting Strict mode did not appear to be possible under the legacy protobuf interface. We are using mesoshttp PyPI package.

gisjedi:
We must ensure that master failover is properly handled by our use of the mesoshttp package.

gisjedi:
Short-term we are rolling the mesoshttp tarball into source, but once they release 0.3.0, we need to update our requirements to consume from official source. Ref: https://github.com/osallou/python-mesos-http/pull/6#issuecomment-436571556

gisjedi:
https://github.com/osallou/python-mesos-http/releases/tag/0.3.0 out so #1415 is OBE

PR #1405, opened 2018-11-01
Strict mode support

gisjedi:
This PR resolves the following issues:

- #1040: Support for Strict mode required an update to the v1 API.
- #1383: Support for Mesos roles was simple to add as part of this effort


gisjedi:
Known problems:

- UI is showing Master and Scheduler as offline... not presently sure why
- Scheduler v4 API will __not__ work in Strict security clusters as the Mesos API calls do not work. Since we are nearing v4 deprecation, I'm okay with this.

PR #1413, opened 2018-11-06
Fixed code that was incorrectly processing SQS messages

jeremysnyder:
In testing this against actual S3 and SQS messages, it appears the SQS message format is different from how this was originally coded.  Making this change made the monitor work with standard SQS S3 notification event messages.


mikenholt:
Thanks for the PR! 👍  

We are in the middle of trying to push a release for a customer.  Once we get that tested and out the door we will take some time to look over this and let you know.  It looks good at first glance, but we may have some additional testing on our end that we need to do.

Issue #1431, opened 2018-11-18
Validate configuration against non-seed job types


JohnPTobe:
The new JobConfiguration class currently only validates against seed manifests. This causes a problem in queueing non-seed job types via the new v6 api with v6 job configuration objects.
Add a validate_old method to handle old style interfaces until non-seed job types are all replaced.

JohnPTobe:
Closed in PR #1432

Issue #1430, opened 2018-11-16
Revert V6 Data Schema Changes


JohnPTobe:
The V6 Data Schema was changed to address this issue: #1409
However, the fix really should have been to get the needed information from the seed manifest for the job.
-Change the get_injected_env_vars method in job/data/job_data.py to accept a seed manifest and use that for the multiple flag
-Change the v6 schema back to how it is described here:
http://ngageoint.github.io/scale/docs/rest/v6/data.html
-Add tests to the v6 data to ensure files are actually added when specified according to the schema.

JohnPTobe:
Closed by PR #1432

PR #1432, opened 2018-11-18
1140 v6 queue

JohnPTobe:
Finishing v6 queue changes started by @Fizz11 
Reverting v6 Data schema changes, addressing #1430 
Adding validate_old method, addressing #1431


scale/job/execution/configuration/configurators.py
```diff
@@ -64,7 +68,29 @@ def configure_queued_job(self, job):
...
+                    param = FileParameter(input_dict['name'], media_types, required, True)
+                    interface.add_parameter(param)
+                elif input_dict['type'] == 'property':
+                    interface.add_parameter(JsonParameter(input_dict['name'], 'string', required))
+
+        env_vars = {}
+        if isinstance(data, JobData):
+            env_vars = data.get_injected_env_vars(input_files_dict, interface)
+        else:
+            env_vars = data.get_injected_env_vars(input_files_dict)
```
emimaesmith: Assuming this calls the `JobData.get_injected_env_vars`, I would expect an error to be thrown here since the method signature doesn't define a default value for `interface`.

PR #1465, opened 2018-12-19
Issue 1429 - purge_recipe unexpected function return

mikenholt:
##### Checklist

- [x] `manage.py test` passes
- [x] tests are included

### Affected app(s)

- recipe
### Description of change

- Added a simple `is_real_node` check when gathering recipe nodes that need to be purged. 

- Full disclosure: I was not able to reproduce this error with just a Scale test.  I have no idea how a DummyNode came into existence with a `node_type` that equals `JobNodeDefinition.NODE_TYPE`

### Background 
https://github.com/ngageoint/scale/blob/8d8cca814936528a333374e35af5122e652f4900/scale/recipe/messages/purge_recipe.py#L94

is getting back `DummyNodeInstance` when it is expecting only `JobNodeInstance`.

Error (actual error is on Line # 96):
```AttributeError: 'DummyNodeInstance' object has no attribute 'job' ```

Also may apply to:
https://github.com/ngageoint/scale/blob/8d8cca814936528a333374e35af5122e652f4900/scale/recipe/messages/purge_recipe.py#L102

### DIFF
--- a/scale/recipe/test/messages/test_purge_recipe.py
+++ b/scale/recipe/test/messages/test_purge_recipe.py
@@ -321,7 +321,7 @@ class TestPurgeRecipe(TransactionTestCase):
             source_file_id=file_2.id), 0)
 
     def test_execute_dummy_recipe(self):
-        """Tests callign PurgeRecipe.execute() with dummy nodes"""
+        """Tests calling PurgeRecipe.execute() with dummy nodes"""
 
         file_2 = storage_test_utils.create_file(file_type='SOURCE')
         trigger = trigger_test_utils.create_trigger_event()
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: scale/recipe/test/messages/test_purge_recipe.py [hunk 1]
```diff
@@ -314,3 +319,70 @@ def test_execute_force_stop(self):
         # Check results are accurate
         self.assertEqual(PurgeResults.objects.values_list('num_recipes_deleted', flat=True).get(
             source_file_id=file_2.id), 0)
+
+    def test_execute_dummy_recipe(self):
+        """Tests callign PurgeRecipe.execute() with dummy nodes"""
```
JohnPTobe: calling this typo out

mikenholt: Good...🕶...call

Issue #1482, opened 2019-01-23
Add ID to job type list REST API results


emimaesmith:
The v6 GET Job Type REST API currently does not return the job id in the list of results. The id should be included in the job-types GET REST API as this information is required for other calls (such as the /metrics API).

emimaesmith:
This change isn't feasible due to the v6 /job-types/ GET REST API returning essentially a summary of each job type listing the latest and all of the different versions available. Each version of the job-type will have its own id, therefore listing a single id for each result will not be adequate. Deferring this addition for future implementation consideration.

emimaesmith:
Requirement satisfied via the addition of the /job-type-names endpoint. See issue #1503 and PR #1509

PR #1484, opened 2019-01-23
#1482 Adding ID to v6 /job-type/ GET REST API results

emimaesmith:
##### Checklist

- [x] `manage.py test` passes
- [x] tests are included

### Affected app(s)

job
### Description of change

Updated the v6 /job-types/ GET REST API call to include the id of the job type in the results:

```
{
        "count": 1,
        "next": null,
        "previous": null,
        "results": [
            {
                "id": 3,
                "name": "my-job",
                "title": "My Job",
                "description": "A simple job type",
                "icon_code": "f013",
                "versions": ["1.0.0"],
                "latest_version": "1.0.0"
            }
        ]
 }
```

### DIFF
--- a/scale/job/job_type_serializers.py
+++ b/scale/job/job_type_serializers.py
@@ -40,7 +40,7 @@ class JobTypeBaseSerializerV6(ModelIdSerializer):
 
 class JobTypeListSerializerV6(JobTypeBaseSerializerV6):
     """Converts job type model fields to REST output"""
-    id = None
+    id = serializers.IntegerField()
     version = None
     versions = serializers.ListField(child=serializers.CharField())
     latest_version = serializers.CharField(source='version')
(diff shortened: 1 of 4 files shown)


scale/job/job_type_serializers.py
```diff
@@ -40,7 +40,7 @@ class JobTypeBaseSerializerV6(ModelIdSerializer):
 
 class JobTypeListSerializerV6(JobTypeBaseSerializerV6):
     """Converts job type model fields to REST output"""
-    id = None
+    id = serializers.IntegerField()
```
JohnPTobe: You should probably just delete the line for consistency and it'll inherit the id from the model serializer.

emimaesmith:
This change isn't feasible due to the v6 /job-types/ GET REST API returning essentially a summary of each job type listing the latest and all of the different versions available. Each version of the job-type will have its own id, therefore listing a single id for each result will not be adequate. Tabling this addition for future implementation consideration.

Issue #1485, opened 2019-01-24
Transition for Replacing Triggers


emimaesmith:
Provide a transitional period for replacing the trigger system with v6.0.0. This should allow for the current trigger system to still function while they are in the process of being converted.

New workspaces and recipes will be created using the new system while existing triggers will still function.

Related to issue #1162 #1181

emimaesmith:
No changes to the UI will be needed for this transition issue. Strikes and Scan configurations created through the API may contain the new 'recipe' configuration definition.

PR #1486, opened 2019-01-29
#1485 - Transition to Replace Triggers

emimaesmith:
##### Checklist

- [x] `manage.py test` passes
- [x] tests are included
- [x] migrations are included

### Affected app(s)

ingest
queue
recipe

### Description of change

We are in the process of replacing the existing trigger system in Scale v6. In order to provide a transitional release, these changes introduce configuring a target recipe for a Strike or Scan while still allowing the existing system to continue to function.

Changes:
Added an optional (for now - in Scale 6 this will become required) `recipe` object to the Strike and Scan configuration. This configuration matches the recipe name, as well as the data/media types of each  input so the recipe may be queued directly from file ingest. 

Also included is a fix for #1490


emimaesmith:
Parse jobs are more complicated as they are kicked off when the scale file is saved. If we want to keep a transition period, we would need a way to tell if the file has been created by a standalone parse job or as part of a recipe. I am currently depending on the job_exe.recipe field to be populated if the parse job is coming from a recipe.

scale/recipe/definition/json/definition_v6.py
```diff
@@ -31,6 +31,7 @@
             'type': 'string',
         },
         'input': INTERFACE_SCHEMA,
+        # 'input_conditon': DATA_FILTER_SCHEMA,
```
JohnPTobe: Why are the filters commented out?

emimaesmith: That's a leftover that didn't get removed from my first idea of implementing this. Originally thought we would have needed conditions for each input, but this is not actually the case. I'll remove the commented out line.

scale/job/serializers.py
```diff
@@ -6,11 +6,11 @@
 import rest_framework.serializers as serializers
 
 from job.models import Job
-from job.job_type_serializers import JobTypeBaseSerializerV5, JobTypeBaseSerializerV6
-from job.job_type_serializers import JobTypeSerializerV5
-from job.job_type_serializers import JobTypeRevisionBaseSerializer
-from job.job_type_serializers import JobTypeRevisionSerializerV5, JobTypeRevisionSerializerV6
-from job.job_type_serializers import JobTypeRevisionDetailsSerializerV6
+# from job.job_type_serializers import JobTypeBaseSerializerV5, JobTypeBaseSerializerV6
```
JohnPTobe: Can these imports be moved back to the top? If not, we need to delete these commented out lines.

Issue #1494, opened 2019-02-19
Requeueing queued jobs


JohnPTobe:
Requeueing jobs that are already queued leaves them in a limbo status where they can never be removed from the queued jobs page.  The requeued job is processed but the queue model is never updated to remove the originally queued job.

PR #1497, opened 2019-02-20
#1494 - Requeueing queued jobs

JohnPTobe:
##### Checklist
- [x] `manage.py test` passes
- [x] tests are included

### Affected app(s)
- queue

### Description of change
If a job is re-queued when it is already in the queue there will be two copies of the job in the queue.  If one of the copies will never be scheduled (e.g. due to invalid resource requirements) it will sit on the queue forever and there's no easy way to get rid of it.  In this scenario it can be assumed that if a user wants to re-queue a job already in the queue, the job already in the queue can be deleted and replaced with the new one.

### DIFF
--- a/scale/queue/models.py
+++ b/scale/queue/models.py
@@ -349,7 +349,7 @@ class QueueManager(models.Manager):
             queue.queued = when_queued
             queues.append(queue)
 
-        Queue.objects.filter(job__in=job_ids).update(is_canceled=True)
+        self.cancel_queued_jobs(job_ids)
 
         if queues:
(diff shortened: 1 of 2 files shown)


JohnPTobe:
After making the change to delete the job from the queue I've gone through the code and convinced myself that a canceled job will be removed from the queue regardless of resource restrictions and we can mark said queue items as canceled rather than deleting them if that is preferable.

JohnPTobe:
Ready for review. I made the change to cancel instead of delete.

Issue #1511, opened 2019-03-05
Reset unmet_resources flag


JohnPTobe:
If a job type is flagged as lacking resources, stop trying to run the job for awhile.  Once the warning has expired, check the job type again to see if the cluster now has the needed resources.  If it does have the needed resources, reset the flag on the job type.  Alternatively, if a job type has been edited manually and thus the flag was reset, clear the warning and check again.

Issue #1493, opened 2019-02-19
Job type resource warnings/errors


JohnPTobe:
The seed spec allows for arbitrary resource names but scale only supports "cpus", "mem", "disk" and "gpus".
We should warn users when creating/editing job types with resources with other names that these resources will not be supported and give strong warnings/errors if the supported resources aren't specified.

JohnPTobe:
Looking into this a bit further, Scale supports arbitrary resource names and it's hard to know which resource names will be offered by mesos.  It could still be useful to implement this to catch possible typos.

PR #1512, opened 2019-03-05
#1511 - Reset unmet_resources flag

JohnPTobe:
##### Checklist
- [x] `manage.py test` passes
- [x] tests are included

### Affected app(s)
-job
-scheduler

### Description of change
Fixing scheduling logic to skip job types that have been marked as having invalid resource requirements for awhile, and only attempt to schedule them again either after a timeout period or after the job type has been updated and the flag reset.  If a job type is again able to be rescheduled due to the needed resource being offered, unset the flag.  Also, added warning when validating a job type with a nonstandard resource (#1493) and cleaned up some documentation TODOs.

### DIFF
--- a/scale/scheduler/manager.py
+++ b/scale/scheduler/manager.py
@@ -13,7 +13,7 @@ from scheduler.models import Scheduler
 from util.active_warnings import ActiveError, ActiveWarning
 from util.parse import datetime_to_string
 
-CLEANUP_WARN_THRESHOLD = datetime.timedelta(hours=3)
+CLEANUP_WARN_THRESHOLD = datetime.timedelta(hours=1)
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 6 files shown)


scale/scheduler/manager.py
```diff
@@ -13,7 +13,7 @@
 from util.active_warnings import ActiveError, ActiveWarning
 from util.parse import datetime_to_string
 
-CLEANUP_WARN_THRESHOLD = datetime.timedelta(hours=3)
+CLEANUP_WARN_THRESHOLD = datetime.timedelta(hours=1)
```
JohnPTobe: Good change. Clean up those warnings. We may want to make it even shorter. 👍

JohnPTobe: Even better, we may want to look into reporting job types rather than a huge list of job ids for troubleshooting.

Issue #1518, opened 2019-03-12
Data Type Periods


JohnPTobe:
Scale post task does not data type tags that have a period in them.

JohnPTobe:
We should probably remove the VALID_TAG_PATTERN check or relax it to only check for unescapable characters.

PR #1523, opened 2019-03-19
#1518 - Data Type Periods

JohnPTobe:
##### Checklist

- [X] `manage.py test` passes
- [X] tests are included
- [X] migrations are included
- [X] documentation is changed or added

### Affected app(s)
- ingest
- storage

### Description of change
Removed the data tag pattern check for data type tags and changed the field in the database from a CharField to an array of CharFields so we don't care what characters are used for tags and we don't have to do any splitting on a comma separated string.


JohnPTobe:
Ready for review

scale/storage/serializers.py
```diff
@@ -112,7 +91,7 @@ class ScaleFileBaseSerializerV5(ModelIdSerializer):
     media_type = serializers.CharField()
     file_type = serializers.CharField()
     file_size = serializers.IntegerField()  # TODO: BigIntegerField?
-    data_type = DataTypeField()
+    data_type = serializers.ListField(child=serializers.CharField(), source='data_type_tags')
```
emimaesmith: Is there a reason this was left as `data_type` instead of `data_type_tags`?

JohnPTobe: It's pointless as v5 is going away but I kept it the field name the same for the v5 serializer out of habit.

JohnPTobe: The v6 field has the name change.

Issue #1538, opened 2019-03-26
V5 Job Details Serializer Bug


JohnPTobe:
The v5 job details serializer expects that an object of type 'files' will be a list but this is not always the case and this causes an error.  As the legacy job data format is going away, a simple fix would be to check if it is a single object or list and acting accordingly.

Issue #1537, opened 2019-03-25
Legacy Job Data is not populated


JohnPTobe:
Currently there are several bugs when populating recipes with legacy job data.  If a job produces no output or one file when the output type is files (multiple), the output is lost when trying to pass it to the next job. We need to create an empty array or an array with one element when this happens.

Issue #1536, opened 2019-03-25
API urls don't support periods


JohnPTobe:
The v6 recipe-types endpoint doesn't support periods in the url, resulting in not being able to get details for some recipes in the v6 api.  Other apis might have this problem as well.

PR #1539, opened 2019-03-26
#1537 - Legacy Job Data is not populated

JohnPTobe:
##### Checklist

- [x] `manage.py test` passes
- [x] tests are included

### Affected app(s)
- data
- ingest
- job
- product
- recipe

### Description of change
Fix for serializer bug (#1538), a bug with periods in urls (#1536) and a bug causing jobs to be stuck pending in recipes.  A legacy job type can execute successfully without creating required output.  This leaves jobs connected to such outputs stuck in a pending state, unable to be queued.  This fix cancels any such pending jobs and any future jobs that get in this state.  Also, jobs that finish without creating required output are marked as failed instead of completing successfully.

### DIFF
--- a/scale/job/test/messages/test_process_job_input.py
+++ b/scale/job/test/messages/test_process_job_input.py
@@ -395,7 +395,7 @@ class TestProcessJobInput(TransactionTestCase):
                 }
             }
         }
-        job_type_1 = job_test_utils.create_job_type(interface=manifest_1)
+        job_type_1 = job_test_utils.create_seed_job_type(manifest=manifest_1)
         manifest_2 = {
             'seedVersion': '1.0.0',
             'job': {
@@ -418,7 +418,7 @@ class TestProcessJobInput(TransactionTestCase):
                 }
             }
         }
-        job_type_2 = job_test_utils.create_job_type(interface=manifest_2)
+        job_type_2 = job_test_utils.create_seed_job_type(manifest=manifest_2)
         output_dict = {
             'version': '1.0',
             'output_data': [{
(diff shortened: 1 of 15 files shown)


scale/job/test/messages/test_process_job_input.py
```diff
@@ -364,3 +364,98 @@ def test_execute_with_recipe(self):
...
+                'interface': {
+                    'command': '',
+                    'inputs': {'files': [], 'json': []},
+                    'outputs': {
+                        'files': [{'name': 'OUTPUT_A', 'pattern': '*.png'}]
+                    }
+                }
+            }
+        }
+        job_type_1 = job_test_utils.create_job_type(interface=manifest_1)
```
emimaesmith: `job_test_utils.create_seed_job_type` should be used instead of `create_job_type`. `create_job_type` will be removed entirely in v7 and using the seed job method will make it easier to merge later.

scale/job/test/messages/test_process_job_input.py
```diff
@@ -364,3 +364,98 @@ def test_execute_with_recipe(self):
...
+                'interface': {
+                    'command': '',
+                    'inputs': {'files': [{'name': 'INPUT_B', 'multiple': True}]},
+                    'outputs': {
+                        'files': [{'name': 'OUTPUT_B', 'pattern': '*.png'}]
+                    }
+                }
+            }
+        }
+        job_type_2 = job_test_utils.create_job_type(interface=manifest_2)
```
emimaesmith: Same comment as above.

JohnPTobe: Good catch.  I just need to handle failing v5 jobs without required output.  I went down a rabbit hole making sure we were checking this for seed jobs and realized at the end we were already doing that correctly.  So now I'm going down the path for legacy jobs.

Issue #1567, opened 2019-04-05
Duplicate Error Names Invalid


JohnPTobe:
Currently the v6 job type validation allows for duplicate error names.  Duplicate error names are invalid in the seed spec and we can't create two errors with the same name/job-type-name combination.

Issue #1566, opened 2019-04-04
Cannot Edit Legacy Job Type Used in V6 Recipe


JohnPTobe:
If a legacy job type is used inside a v6 recipe type it cannot be edited.  The edit_job_type_v5 call attempts to validate against recipe definitions and v6 recipe definitions don't have the correct method.

Issue #1565, opened 2019-04-04
Input Metadata is too long


JohnPTobe:
The input metadata environment variable is too long for some jobs in scale.  Jobs with 1000+ input files result in a json string of over 180000 characters.  We need to possibly create an input_metadata manifest file with all the file urls.

Issue #1564, opened 2019-04-03
Workspace Lost When Superseding V6 Recipe


JohnPTobe:
Superseding legacy recipes was fixed, but not v6 recipes.  We need to copy the recipe_config when superseding v6 recipes.

PR #1573, opened 2019-04-10
#1564 - Workspace Lost When Superseding V6 Recipe

JohnPTobe:
##### Checklist

- [x] `manage.py test` passes
- [x] tests are included

### Affected app(s)
- job
- recipe

### Description of change
Fixes for bugs found while testing 5.9.4:
#1564, #1566, #1567, #1565 
The first three bugs are relatively straightforward. The last one with really long (180k characters) input metadata requires changing the way we pass input metadata to creating an input manifest file in the pre step instead of when configuring the job.  If a job interface specifies an input metadata manifest as an input we create an input metadata dictionary and convert it to a json manifest and put it in the input workspace.

### DIFF
--- a/scale/job/management/commands/scale_pre_steps.py
+++ b/scale/job/management/commands/scale_pre_steps.py
@@ -54,6 +54,8 @@ class Command(BaseCommand):
             logger.info('Validating outputs and workspaces...')
             job_interface.validate_workspace_for_outputs(exe_config)
 
+            self._generate_input_metadata(job_exe)
+
             job_data = job_exe.job.get_job_data()
             job_data = JobDataSunset.create(job_interface, job_data.get_dict())
             logger.info('Setting up input files...')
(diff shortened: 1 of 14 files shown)


scale/job/management/commands/scale_pre_steps.py
```diff
@@ -82,3 +89,96 @@ def _get_job_exe(self, job_id, exe_num):
...
+        :type job_exe: :class:`job.models.JobExecution`
+        :returns: The remote path for storing the manifest
+        :rtype: str
+        """
+
+        remote_path = ''
+        if job_exe.job.recipe:
+            recipe = job_exe.job.recipe
+            recipe_type_path = get_valid_filename(recipe.recipe_type.name)
+            recipe_version_path = get_valid_filename(recipe.recipe_type.version)
```
emimaesmith: Weren't recipe type versions removed in favor of revisions?

Issue #1580, opened 2019-04-11
Batches Taking Too Long


JohnPTobe:
There seems to be a bug with batches creating either a message bomb and/or taking too long to run.  Temporarily disable the batches api until this potential bug is addressed.

PR #1592, opened 2019-04-17
#1580 - Batches Taking Too Long

JohnPTobe:
##### Checklist

- [x] `manage.py test` passes
- [x] tests are included

### Affected app(s)
- recipe

### Description of change
Fixing message bomb caused when calling create recipe messages with over 100 recipes.  The message variable was never reset, so if you had 101 recipes you ended up with two messages with the same 101 recipes in both.


JohnPTobe:
This will also close #1593

Issue #1598, opened 2019-04-23
Remove Required Recipe Type Revision Number From Ingest Configuration 


emimaesmith:
Currently the ingest configuration requires a recipe object to be specified that contains a recipe type name and revision number. The revision number field should be moved to an optional parameter which, if not specified, would be defaulted to the latest revision of the recipe. When the parse_recipe_handler is being processed, if a revision number is specified, the RecipeTypeRevision table should be queried instead of the RecipeType table in order to retrieve the proper RecipeType object.

emimaesmith:
This will fix #1598

emimaesmith:
Closed via #1602

Issue #1594, opened 2019-04-17
Updating Recipe Revision Number Breaks Strikes


emimaesmith:
Strikes require a recipe name and revision number. If the recipe is updated, the strike is not updated with the latest revision number and the ingest job will throw an error about the recipe type not existing.

emimaesmith:
Closed via #1602

Issue #1591, opened 2019-04-17
Recipe set_recipe_input_data_v6 Contains Trigger References


emimaesmith:
The recipe.models.RecipeManager.set_recipe_input_data_v6 method references removing LOC when triggers are removed. Need to understand the functionality behind why this is trigger related and any consequences/changes that will need to be made when the marked lines are removed.

```
def set_recipe_input_data_v6(self, recipe, input_data):
        """Sets the given input data as a v6 JSON for the given recipe. The recipe model must have its related
        recipe_type_rev model populated.

        :param recipe: The recipe model with related recipe_type_rev model
        :type recipe: :class:`recipe.models.Recipe`
        :param input_data: The input data for the recipe
        :type input_data: :class:`data.data.data.Data`

        :raises :class:`data.data.exceptions.InvalidData`: If the data is invalid
        """

        # TODO: remove when legacy trigger system is removed
        definition_version = recipe.recipe_type_rev.definition['version']

        recipe_definition = recipe.recipe_type_rev.get_definition()
        input_data.validate(recipe_definition.input_interface)
        input_dict = None

        # TODO: this code path supports passing output workspace ID in recipe data, remove when legacy trigger system is
        # removed
        config = None
        if definition_version == '1.0' and recipe.recipe:
            if 'workspace_id' in recipe.recipe.input:
                input_dict = convert_data_to_v1_json(input_data, recipe_definition.input_interface).get_dict()
                input_dict['workspace_id'] = recipe.recipe.input['workspace_id']

                # store workspace id in config so it isn't lost
                workspace_id = input_dict['workspace_id']
                workspace = None
                try:
                    workspace = Workspace.objects.get(pk=workspace_id)
                except Workspace.DoesNotExist:
                    logger.exception('Could not set workspace from input data. Workspace does not exist: %d', workspace_id)

                config = RecipeConfigurationV6(recipe.configuration)
                if workspace:
                    config.get_configuration().default_output_workspace = workspace.name
                    config = config.get_dict()

        if not input_dict:
            input_dict = convert_data_to_v6_json(input_data).get_dict()

        if config:
            self.filter(id=recipe.id).update(input=input_dict, configuration=config)
        else:
            self.filter(id=recipe.id).update(input=input_dict)
```

emimaesmith:
Closed via #1602

Issue #1508, opened 2019-03-01
Move Scale to v7


emimaesmith:
The next Scale version will be 7.0 instead of 6.0. Functionality will be identical to v6. Updates to the API will need to be made to ensure we support v7. We can do this two ways:

1. Replace all v6 API checks with v7
2. Support both v6 and v7 calls.

Supporting both v6 and v7 calls would uphold supporting one version behind. v6 and v7 would perform exactly the same.

We also need to decide what to do with the v6 schemas - do we add additional v7 schemas? Do we leave the schema versions as v6 for a Scale 7.0?

emimaesmith:
Added v7 to all API checks in #1602

PR #1602, opened 2019-04-29
#1598 #1594 #1591 #1508

emimaesmith:
##### Checklist

- [x] `manage.py test` passes
- [x] tests are included

### Affected app(s)

batch, diagnostic, error, ingest, job, metrics, node, queue, recipe, scheduler, storage

### Description of change

#1598 Removed the requirement for the recipe revision number from the strike/scan configuration. Revision number is optional and the latest will be used if not specified.
#1594 Fixed bug where updating a recipe type (causing an updated revision number) broke the strike/scan configuration due to a recipe not found error
#1591 Removed leftover old trigger references
#1508 Added v7 API calls to the views


emimaesmith:
Ready for review. I can include specific tests for v7 API calls, but since functionality is exactly the same as v6, I didn't think it was completely necessary to duplicate test cases.

scale/diagnostic/views.py
```diff
@@ -23,7 +23,16 @@ class QueueScaleBakeView(GenericAPIView):
...
+    
+    def get_serializer_class(self):
+        """Returns the appropriate serializer based off the requests version of the REST API. """
+
+        if self.request.version == 'v6':
+            return QueueStatusSerializerV6
+        elif self.request.version == 'v7':
+            return QueueStatusSerializerV6
+        else:
+            return QueueStatusSerializer
```
JohnPTobe: We should get rid of this else and remove the queuestatusserializer class

scale/diagnostic/views.py
```diff
@@ -76,7 +87,16 @@ class QueueScaleCasinoView(GenericAPIView):
...
+
+    def get_serializer_class(self):
+        """Returns the appropriate serializer based off the requests version of the REST API. """
+
+        if self.request.version == 'v6':
+            return QueueStatusSerializerV6
+        elif self.request.version == 'v7':
+            return QueueStatusSerializerV6
+        else:
+            return QueueStatusSerializer
```
JohnPTobe: same as above

scale/diagnostic/views.py
```diff
@@ -129,7 +151,16 @@ class QueueScaleHelloView(GenericAPIView):
...
```
JohnPTobe: same

scale/queue/views.py
```diff
@@ -50,6 +50,10 @@ def get_serializer_class(self):
 
         if self.request.version == 'v6':
             return QueueStatusSerializerV6
+        elif self.request.version == 'v7':
+            return QueueStatusSerializerV6
+        else:
+            return QueueStatusSerializer
```
JohnPTobe: remove queuestatusserializer class

scale/diagnostic/views.py
emimaesmith: Removed

scale/diagnostic/views.py
emimaesmith: Removed

scale/queue/views.py
emimaesmith: Removed

scale/diagnostic/views.py
emimaesmith: Removed

Issue #1601, opened 2019-04-25
Reprocessing Recipe Doesn't Use Latest Recipe Revision


emimaesmith:
Reprocessing a recipe will not pull the latest revision of the recipe when it is reprocesses. By default it should use the latest revision, but also provide the option to specify a particular revision. 

Example:
Jobs 1 and 2 in Recipe A were updated to change the names of the inputs and outputs. Recipe A's definition was not changed. Jobs are now failing due to mismatched input/outputs. Recipe A was then updated, and previous recipes are re-processed. Jobs are still failing because the reprocessed jobs are still using the old recipe revision and not the fixed one.

PR #1607, opened 2019-05-01
#1601 Reprocessing Recipe Doesn't Use Latest Revision

emimaesmith:
##### Checklist

- [x] `manage.py test` passes
- [x] tests are included

### Affected app(s)

recipe

### Description of change

The recipe reprocessing endpoint was updated to always pull the latest revision of the recipe type, unless 'revision_num' is specified in the incoming json data. If the revision number is not found a 404 is returned (same as if recipe id not found).

### DIFF
--- a/scale/recipe/test/test_views.py
+++ b/scale/recipe/test/test_views.py
@@ -1375,38 +1375,6 @@ class TestRecipeReprocessViewV6(APITransactionTestCase):
         self.recipe_type = recipe_test_utils.create_recipe_type_v6(name='my-type', definition=def_v6_dict)
         self.recipe1 = recipe_test_utils.create_recipe(recipe_type=self.recipe_type, input=self.data, config=self.config)
         recipe_test_utils.process_recipe_inputs([self.recipe1.id])
-        
-        # Create job with two revisions
-        def_rev_v6_dict = {'version': '6',
-                       'input': {'files': [{'name': 'INPUT_FILE', 'media_types': ['image/tiff'], 'required': True,
-                                            'multiple': True}],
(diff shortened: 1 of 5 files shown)


scale/recipe/test/test_views.py
```diff
@@ -5,6 +5,7 @@
 import django
 import json
 
+from django.db import transaction
```
JohnPTobe: unnecessary import?

emimaesmith: Yep. Removed

scale/recipe/views.py
```diff
@@ -631,9 +632,16 @@ def _post_v6(self, request, recipe_id):
             raise BadParameter(unicode(ex))
 
         try:
-            recipe = Recipe.objects.select_related('recipe_type', 'recipe_type_rev').get(id=recipe_id)
+            recipe = Recipe.objects.select_related('recipe_type').get(id=recipe_id)
+            if revision_num:
+                recipe.recipe_type_rev = RecipeTypeRevision.objects.get_revision(recipe.recipe_type.name, revision_num)                
+            else:
+                revision_num = recipe.recipe_type.revision_num
+                recipe.recipe_type_rev = RecipeTypeRevision.objects.get_revision(recipe.recipe_type.name, recipe.recipe_type.revision_num)
```
JohnPTobe: Add comment to clarify that we're checking if the revision exists and use a throwaway variable to store the result rather than making it look like we're updating the recipe we got with a new revision object. Also, I don't think we need to check if the revision number exists if we're using the existing recipe type's revision_num variable.

Issue #1626, opened 2019-05-13
Legacy Job Types Crash Scale


JohnPTobe:
Legacy Job Types crash Scale on startup and potentially other places with an InvalidSeedManifestDefinition exception.  We need to guard against such exceptions as we can't count on legacy job types being cleaned up. We also probably want to have a way to handle legacy job types for showing old jobs, etc.

PR #1629, opened 2019-05-16
#1626 - Legacy Job Types Crash Scale

JohnPTobe:
### Checklist

- [x] `manage.py test` passes
- [x] migrations are included

### Affected app(s)
- job
- scheduler

### Description of change
Adding migrations to update legacy job types before v5 code is removed and updater code to update job types that were missed (perhaps if a database was migrated before this fix).


JohnPTobe:
Only if someone manually un-pauses a legacy job type.  We could go the extra step of not allowing jobs to even be created for legacy job types.  We could delete them but I'd prefer to keep them around so old jobs and other models aren't left dangling pointing at nothing.

PR #1635, opened 2019-05-22
Prerelease 7.0.0 cleanup

gisjedi:
##### Checklist

- [x] `manage.py test` passes
- [x] tests are included

### Affected app(s)

* ingest
* job
* metrics
* recipe
* queue

### Description of change
* Removed fixture that was causing problems across scheduler restarts. This will have to be resolved in the future
* Added defaults to the Task base class to fix tasks not launching in Docker
* Replaced statement to add all docker parameters that was lost in v5 cleanup
* Fixed version of clock and strike system jobs
* Fixed issue with the diagnostic recipe types fixture not installing properly
* Fixed references to removed queue_new_jobs methods


scale/diagnostic/fixtures/diagnostic_recipe_types.json
```diff
@@ -1,46 +0,0 @@
...
-                                          "node_type": {"node_type": "job", "job_type_name": "scale-roulette", "job_type_version": "1.0.0", "job_type_revision": 1}}}
-            },
-            "revision_num": 1,
-            "created": "2017-11-13T00:00:00.0Z",
-            "last_modified": "2017-11-13T00:00:00.0Z"
-        }
-    },
-    {
-        "model": "recipe.RecipeTypeRevision",
-        "pk": null,
```
JohnPTobe: Could the problem be the revision?  The model code creates the recipe type and it's first revision and then attempts to create the revision again with this fixture?

emimaesmith: Problem was with the RecipeType model not defining a `natural_key` method that returned a tuple with just the recipe_type.name value `return (self.name,)` 😑

Issue #1149, opened 2018-04-20
Clear out database update


dfaller:
Clear out all of the updates in the database update system task. Users will need to have completed all of these updates in the most recent v5 release before moving to v6. Once Scale is at v6, we can assume that all of these updates have been completed.

Apply any needed database migrations that can occur now that the updates have been removed. For example, add a job_exe index (job_id, exe_num) and remove batch_job and batch_recipe models.

JohnPTobe:
The database updates have been removed, we can add migrations to clean up the database accordingly.

JohnPTobe:
What would we add a job_exe index for?  What would it do?

JohnPTobe:
Is it related to #736?

JohnPTobe:
There is already an index on job_exe but there are fields that were moved to the output and end models so they should be removed. Also, we should open another issue investigating the index being reversed in aws's rds (exe_num, job_id)

Issue #1055, opened 2017-12-22
Change job execution index to be unique


dfaller:
Change the job execution index (combining job ID and exe num) to be unique. This will prevent duplicate job execution models from being created.

We've had some situations where two schedulers run simultaneously and create duplicate job executions.

This might be tricky to do. We need to ensure that there are no duplicates before adding the new index.

PR #1655, opened 2019-06-03
#1149 - Clear out database update

### LINKED ISSUE #1642: Stale Status
The status api in our dev cluster returns no content because the timestamp is over 12 seconds old (between 30-40).  We need to either fix whatever is slowing down status generation or increase the freshness threshold.

JohnPTobe:
##### Checklist

- [x] `manage.py test` passes
- [x] tests are included
- [x] migrations are included

### Affected app(s)
- batch
- job
- metrics
- product
- recipe
- scheduler
- storage

### Description of change
Removed BatchJob and BatchRecipe models and changed the job,exe_num index on job_exe tables to be unique.  Also fixed #1642.  I also removed job_exe fields that have been moved to the output and end tables.


scale/job/models.py
```diff
@@ -1666,7 +1638,7 @@ def set_cluster_id(self, framework_id, job_id, exe_num):
     class Meta(object):
         """Meta information for the database"""
         db_table = 'job_exe'
-        index_together = ['job', 'exe_num']
+        unique_together = ['job', 'exe_num']
```
emimaesmith: Is there a reason for taking out the indexing? This could be the index the issue was referring to.

scale/product/migrations/0008_auto_20170221_1413.py
```diff
@@ -12,15 +12,15 @@ class Migration(migrations.Migration):
 
     def populate_file_ancestry_link_batch(apps, schema_editor):
         # Go through all of the BatchRecipe models and update the corresponding FileAncestryLink models
-        BatchRecipe = apps.get_model('batch', 'BatchRecipe')
+        Recipe = apps.get_model('recipe', 'Recipe')
```
emimaesmith: Is this function necessary now that BatchRecipe has been removed? I'm assuming this is to not lose any file ancestry that was once stored in the BatchRecipe?

scale/job/models.py
JohnPTobe: I'm going off of this article that says postgres creates an index when adding the constraint.
https://medium.com/flatiron-engineering/uniqueness-in-postgresql-constraints-versus-indexes-4cf957a472fd 
So the index was already there, I think derick might have meant making it unique, which makes sense as that constraint could be applied once the database update was no longer needed.

scale/product/migrations/0008_auto_20170221_1413.py
JohnPTobe: You're right, but do we delete the whole migration then?  We'd have to update the migrations that follow it and make sure they don't get reapplied.  Maybe it's simpler than I realize.

JohnPTobe: Tossed it in the fire

scale/scheduler/sync/job_type_manager.py
```diff
@@ -80,10 +80,12 @@ def sync_with_database(self):
         updated_job_types = {}
         for job_type in JobType.objects.all().iterator():
             try:
+                job_type.title = job_type.get_title()
+                job_type.description = job_type.get_description()
                 updated_job_types[job_type.id] = job_type
                 update_job_type_resources.append(job_type.get_resources())
             except InvalidSeedMetadataDefinition as ex:
-                logger.info('Invalid Seed manifest for job type %s-%s, id=%d' % (job_type.name, job_type.version, job_type.id))
+                logger.info('Invalid Seed manifest for job type %s-%s, id=%d\n Exception: %s' % (job_type.name, job_type.version, job_type.id, ex))
```
gisjedi: I'd recommend changing this to a `logger.exception` from `logger.info`. That eliminates the need to manually inject the exception and will provide the entire stack trace rather than just the exception name. Its a lot more useful.

JohnPTobe:
This will also address #1055, $1658 and #1659

scale/scheduler/sync/job_type_manager.py
JohnPTobe: Done

Issue #1634, opened 2019-05-21
Attach Condition to Optional Output


JohnPTobe:
Conditions should be able to be connected to optional outputs in recipes, allowing for child jobs to be run or not run based on whether the optional output exists.

PR #1657, opened 2019-06-04
#1634 - Attach Condition to Optional Output

JohnPTobe:
### Checklist
- [x] `manage.py test` passes
- [x] tests are included

### Affected app(s)
- job
- recipe

### Description of change
Actually copy the input interface of a condition and update the copy so we're not modifying the original input interface and making everything required.  Also fixed a bug in setting the json output for a job in the post steps


JohnPTobe:
Decided I should add a test for getting json outputs

Issue #1042, opened 2017-12-18
Return all JSON fields in correct version for API called


dfaller:
Update all of the detail serializers to make sure that all of the JSONs returned are returned in the correct version for the API called.

PR #1663, opened 2019-06-10
#1508 - Support Configuration Version 7

emimaesmith:
##### Checklist

- [x] `manage.py test` passes
- [x] tests are included

### Affected app(s)

batch
data
ingest
job
metrics
queue
recipe
storage 
### Description of change

#1508  
Added `7` as valid configuration version in order to support Scale v7. Update configuration schema version checks to accept both 6 or 7 as the proper version.


emimaesmith:
This should also cover #1042

scale/job/configuration/json/job_config_v6.py
```diff
@@ -156,11 +156,7 @@ def __init__(self, config=None, existing=None, do_validate=False):
         if 'version' not in self._config:
             self._config['version'] = SCHEMA_VERSION
 
-        if self._config['version'] == '2.0':
-            # v2 and v6 are identical
-            self._config['version'] = SCHEMA_VERSION
```
JohnPTobe: This was added back in because we were getting 500 errors for job details pages, the job types api, etc.  We need to either keep this or add calls to convert schemas.

emimaesmith: Check has been added back in.

scale/ingest/test/triggers/test_ingest_recipe_handler.py
```diff
@@ -38,7 +38,7 @@ def setUp(self):
         self.source_file.add_data_type_tag('type3')
         manifest = job_test_utils.create_seed_manifest(inputs_files=[{'name': 'INPUT_FILE', 'media_types': ['text/plain'], 'required': True, 'multiple': True}], inputs_json=[])
         self.jt1 = job_test_utils.create_seed_job_type(manifest=manifest)
-        recipe_type_def = {'version': '6',
```
gisjedi: Shouldn't we duplicate the tests for 6 and 7 so that we are confident we have consistency between them? At some point 7 will drift on past what 6 supports.

emimaesmith: Added another type definition to test both 6 and 7.

emimaesmith:
Closes #1042

PR #1675, opened 2019-06-24
Bug fixes

mikenholt:
##### Checklist

- [x] `manage.py test` passes
- [ ] tests are included
- [ ] migrations are included

### Affected app(s)

- diagnostic
- recipe
- queue
- job

### Description of change

- Wrapped all diagnostic jobs in recipes.
- Fixed bug where a job node with 0 outputs would block children nodes from running.
- Forced the `recipe.pk` into `root_recipe_id` if there is no root or superseded recipe.  If `root_recipe_id` continues on as `None` a recipe will not complete. 
- Fixed a bug where `calculate_topological_order` was doing calculations in place and not persisting the values it was gathering. 
- Fixed a logic check on determining if a `DummyNode` can turn into a real node.  `self.parental_acceptance[name]` is bool that defines if a child node should be blocked from creation based off its status. `parent_node.is_accepted()` is a bool that defines if the parent node has been accepted.  
- Removed reference to field that was remove when trying to get logs.

### DIFF
--- a/scale/diagnostic/fixtures/diagnostic_recipe_types.json
+++ b/scale/diagnostic/fixtures/diagnostic_recipe_types.json
@@ -1,5 +1,5 @@
 [
-   {
+    {
         "model": "recipe.RecipeType",
         "pk": null,
         "fields": {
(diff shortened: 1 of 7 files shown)


gisjedi:
This is a procedural comment. You've fixed a lot of recipe related issues. I'm not sure its required to create issues for all of them, but we probably want to at least capture some of effort required in the MSR.

mikenholt:
> This is a procedural comment. You've fixed a lot of recipe related issues. I'm not sure its required to create issues for all of them, but we probably want to at least capture some of effort required in the MSR.

I broke it out more in DevOps 😉

Issue #1645, opened 2019-05-29
v6 batches don't override priority


wong-j:
Submitting a v6 batch job using a configuration priority on the batch will not override the priority value of the job-type.  It will get inserted into the queue table as the default job-type priority.

JohnPTobe:
I was finally able to reproduce this bug.  Jobs created from legacy job types will get the priority of the job type and that takes precedence over everything in the queue_jobs method.  As legacy job types have been removed along with the priority field of jobs, this is no longer an issue.

PR #1694, opened 2019-07-03
#1645 - v6 batches don't override priority

JohnPTobe:
##### Checklist

- [x] `manage.py test` passes
- [x] tests are included

### Affected app(s)
- batch
- job

### Description of change
Some tests I wrote when trying to reproduce #1645. This will ensure we don't break this in the future.


JohnPTobe:
:smirk:

Issue #1689, opened 2019-07-02
Workspace Description Optional


Rachel-ais:
The Workspace description field is optional in the documentation but when creating the workspace it will return and error. Although the workspace will validate correctly. 

This is the Error: 
```
ERROR [root(21)] Exception handling request for /api/v6/workspaces/.
...
   File "/opt/scale/storage/views.py", line 259, in create
     return self._create_v6(request)
   File "/opt/scale/storage/views.py", line 291, in _create_v6
     workspace = Workspace.objects.create_workspace(name, title, description, configuration, base_url, is_active)
   File "/opt/scale/storage/models.py", line 868, in create_workspace
     workspace.save()
...
 IntegrityError: null value in column "description" violates not-null constraint
 DETAIL:  Failing row contains (3, untitled-workspace, Untitled Workspace, null, null, t, {"broker": {"type": "host", "host_path": "eretete", "credentials..., 2019-07-02 16:26:10.361011+00, null, 2019-07-02 16:26:10.361039+00, t).
```

mheppner:
Rachel, you can ignore this error for now by supplying the `"description"` field, whether it's `null` or an empty string.

I'm not sure the correct way to fix this though. The [`parse_string()`](https://github.com/ngageoint/scale/blob/e3c47c99189a1dd8a3e6ff6cc3b5c50b1f2039da/scale/util/rest.py#L229) function has `default_value` set to `None` and the function will happily return `None` if `required=False`. The docstring indicates `default_value` should be a string, and that the return value should also be string, but both of these conditions are not forced in the code. To complicate this confusion, [`test_parse_string_optional()`](https://github.com/ngageoint/scale/blob/e3c47c99189a1dd8a3e6ff6cc3b5c50b1f2039da/scale/util/test/test_rest.py#L162) explicitly tests if `parse_string()` returns `None`.

This can be fixed by at least checking if the parsed value is in fact a string:

```python
def parse_string(request, name, default_value=None, required=True, accepted_values=None):
    value = _get_param(request, name, default_value, required)
    if not isinstance(value, basestring):
        raise BadParameter('Parameter must be a valid string: "%s"' % name)
    ...
```

Which will break the test that's checking for a `None` return value.

mheppner:
The extent of this issue is pretty broad, `parse_string(required=False)` is used approximately 75 times throughout the serializers, `/api/v6/workspaces/` is not the only endpoint affected by this.

emimaesmith:
[...]

PR #1707, opened 2019-07-11
#1689 - Workspace Description Optional

emimaesmith:
##### Checklist

- [x] `manage.py test` passes
- [x] migrations are included

### Affected app(s)

storage

### Description of change

Documentation lists the workspace description field as optional but an IntegrityError is thrown if a Null value is provided to the database. Updated the workspace description field to truly be optional and allow null values in the database.


emimaesmith:
Closes #1689

Issue #1721, opened 2019-07-19
Job Type Edit Endpoint Consistency


emimaesmith:
**Pain Point? Please describe.**
The job-type/ POST and job-type/name/version PATCH allows you change different fields when editing a job type depending on which endpoint you hit. These endpoints need to be consistent during the edit of a job type.

**Desired Solution**
Allow updates to the manifest and docker_image during a patch - will need to verify the manifest name/version matches the job you're trying to update.
Allow updates to the is_active and is_paused on the POST

**Alternative / Workaround**
Use the right endpoint depending on what you want to change.

emimaesmith:
Closed via #1729

PR #1729, opened 2019-07-24
#1721 Job Type POST/PATCH Endpoint Consistency

emimaesmith:
##### Checklist

- [ ] `manage.py test` passes
- [x] tests are included

### Affected app(s)

job

### Description of change

#1721 
Updated the job-type POST and job-type/name/version PATCH endpoints to update the same fields when editing an existing job type. Added the `is_paused` and `is_active` fields to the job-type/ POST, and `docker_image` and `manifest` fields to the PATCH. An error is returned during the PATCH if the name or version from the manifest does not match the name and version of the job type being updated.

### DIFF
--- a/scale/job/models.py
+++ b/scale/job/models.py
@@ -2075,7 +2075,7 @@ class JobTypeManager(models.Manager):
             job_type.docker_image = docker_image
         if icon_code:
             job_type.icon_code = icon_code
-        if is_active and job_type.is_active != is_active:
+        if is_active is not None and job_type.is_active != is_active:
             job_type.deprecated = None if is_active else timezone.now()
             job_type.is_active = is_active
             if auto_update:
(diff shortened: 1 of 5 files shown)


scale/job/models.py
```diff
@@ -2075,15 +2075,15 @@ def edit_job_type_v6(self, job_type_id, manifest=None, docker_image=None, icon_c
             job_type.docker_image = docker_image
         if icon_code:
             job_type.icon_code = icon_code
-        if is_active and job_type.is_active != is_active:
+        if is_active is not None and job_type.is_active != is_active:
             job_type.deprecated = None if is_active else timezone.now()
```
JohnPTobe: good catch. apparently we couldn't set a job type to inactive before

scale/job/models.py
```diff
@@ -2075,15 +2075,15 @@ def edit_job_type_v6(self, job_type_id, manifest=None, docker_image=None, icon_c
...
             job_type.is_active = is_active
             if auto_update:
                 recipe_ids = RecipeTypeJobLink.objects.get_recipe_type_ids([job_type.id])
                 msgs = [create_activate_recipe_message(id, is_active) for id in recipe_ids]
                 CommandMessageManager().send_messages(msgs)
                 
-        if is_paused and job_type.is_paused != is_paused:
+        if is_paused is not None and job_type.is_paused != is_paused:
```
JohnPTobe: same with pausing

Issue #1718, opened 2019-07-17
Deprecation Warning


JohnPTobe:
**Pain Point? Please describe.**
When deprecating a job type and/or recipe type, recipe types that contain them are automatically deprecated (made inactive) if auto-update is set to true.  Users are not warned of this fact, however and it's hard to know which recipe types will be affected.  Scale should display a warning that this will occur and list affected recipe types so the user can correct them if needed.

**Desired Solution**
When a job type or recipe type is deprecated and auto-update is set to true, display a warning listing recipe types that will be affected.  This can be returned via a response from the patch api similar to validation calls or it can be a system warning. If the latter, it can either timeout or be updated each time a job type or recipe type is edited until all affected recipe types are cleared up and there are no more active recipe types with inactive nodes.

**Alternative / Workaround**
Currently there is no easy way to see a list of recipe types containing a job type or recipe type.

**Additional Context**

emimaesmith:
Possibly add a field to the recipe type serializer to check if all jobs and sub-recipes are active. The UI would be able to use this field to add an indicator to the recipe type card to display if there is an issue with the recipe type that needs to be addressed (lists the number of inactive job/recipe types in the recipe-type list view card?). Similar indicator can be added to the job type in the recipe type view to visually indicate if the job type is inactive.

JohnPTobe:
This relates to #1609

JohnPTobe:
The recipe type details api essentially does this with the job_types and sub_recipe_types fields which have the is_active flag for each job/recipe type.  This isn't available when listing all recipe types and I think it would make that api too slow.  We need to decide if we want to return something from the edit method or just display a general warning that recipes may be deprecated along with the current job type or recipe type.

JohnPTobe:
Also I think we should always deprecate recipes that contain a deprecated job/recipe regardless of the auto_update flag.  That flag should be reserved for simply updating the revision of jobs/recipes automatically.

PR #1730, opened 2019-07-28
#1718 - Deprecation Warning

JohnPTobe:
##### Checklist

- [x] `manage.py test` passes
- [x] tests are included

### Affected app(s)
- job
- recipe

### Description of change
Added fields to job-type details and recipe-type details apis to show which recipes each belongs to so they can be easily found if they need to be edited.
Automatically deprecate parent recipes when jobs/recipes are deprecated regardless of auto-update flag. Parent recipes aren't automatically reactivated.
Changed job/recipe patch endpoints to return 200 OK response with validation dict for warnings.

### DIFF
--- a/scale/job/models.py
+++ b/scale/job/models.py
@@ -2075,11 +2075,15 @@ class JobTypeManager(models.Manager):
             job_type.docker_image = docker_image
         if icon_code:
             job_type.icon_code = icon_code
+        deprecated_warning = ""
         if is_active is not None and job_type.is_active != is_active:
             job_type.deprecated = None if is_active else timezone.now()
             job_type.is_active = is_active
             if is_active == False:
                 recipe_ids = RecipeTypeJobLink.objects.get_recipe_type_ids([job_type.id])
+                if recipe_ids:
+                    deprecated_warning = "Recipes were deprecated as a result of deprecating this job type. " \
+                                         "Look at the recipe_types field for recipes that may need to be updated."
                 msgs = [create_activate_recipe_message(id, is_active) for id in recipe_ids]
                 CommandMessageManager().send_messages(msgs)
                 
@@ -2095,6 +2099,8 @@ class JobTypeManager(models.Manager):
         from scheduler.sync.job_type_manager import job_type_mgr
         job_type_mgr.sync_with_database()
 
+        job_type.deprecated_warning = deprecated_warning
+
         # Save any secrets to Vault
         if secrets:
             self.set_job_type_secrets(job_type.get_secrets_key(), secrets)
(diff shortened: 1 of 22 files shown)


JohnPTobe:
It's needs tests and documentation yet but I wanted @emimaesmith and @Rachel-ais to look at this to see if you agree this would work for warning the user about deprecated recipes.  Also, could we use this to display recipe types a job/recipe is part of on the details page?

JohnPTobe:
Ready for review

scale/job/views.py
```diff
@@ -363,6 +369,7 @@ def patch(self, request, name, version):
         except JobType.DoesNotExist:
             raise Http404
 
+        jt = None
```
emimaesmith: Leftover debugging variable?

scale/recipe/models.py
```diff
@@ -1236,8 +1243,13 @@ def edit_recipe_type_v6(self, recipe_type_id, title=None, description=None, defi
         
         if is_active is not None:
             recipe_type.is_active = is_active
-            if auto_update:
+            if is_active == False:
```
emimaesmith: Same question as above about removing the auto_update.

scale/recipe/views.py
```diff
@@ -233,6 +233,7 @@ def patch_v6(self, request, name):
         except RecipeType.DoesNotExist:
             raise Http404
 
+        rt = None
```
emimaesmith: Leftover debugging variable?

scale/job/models.py
```diff
@@ -2078,8 +2082,13 @@ def edit_job_type_v6(self, job_type_id, manifest=None, docker_image=None, icon_c
         if is_active is not None and job_type.is_active != is_active:
             job_type.deprecated = None if is_active else timezone.now()
             job_type.is_active = is_active
-            if auto_update:
+            if is_active == False:
```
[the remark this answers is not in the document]
JohnPTobe: After thinking a bit about it, I think it makes more sense to deprecate all parent recipes when a child is deprecated as they can't be run until the child is fixed.  The auto_update flag should only handle automatically updating the revision of a child job/recipe when it won't break anything.  Also, we obviously don't want to automatically re-enable recipes as there could be other things causing a recipe to be deprecated and a user should manually review a recipe before it's activated.

scale/job/views.py
JohnPTobe: yep, missed a spot

scale/recipe/views.py
JohnPTobe: another spot

JohnPTobe:
Ready for re-review. If you're unsure about the auto_update flag we can discuss it more.

PR #1731, opened 2019-07-29
Issue 1700 - Recipe validation warning if name is not an existing recipe

mikenholt:
##### Checklist

- [x] `manage.py test` passes
- [x] tests are included

### Affected app(s)

- recipe
### Description of change

Added an endpoint that can be used to validate an existing recipe.  With this Scale will know if the recipe that is trying to be validated is existing or not. 

Slight WIP - I still need to update the docs. 

Also, I did not include the same "fix" for jobs as it looks like this type of warning is not present there.  Should we allow for the same `/job-types/{name}/validation` endpoint to work?

### DIFF
--- a/scale/recipe/models.py
+++ b/scale/recipe/models.py
@@ -1341,13 +1341,15 @@ class RecipeTypeManager(models.Manager):
             rt.sub_recipe_types = definition.get_recipe_type_names()
         return recipe_types
 
-    def validate_recipe_type_v6(self, name, definition_dict):
+    def validate_recipe_type_v6(self, name, definition_dict, new_recipe_type):
         """Validates a recipe type prior to attempting a save
 
         :param name: The optional system name of a recipe type being updated
         :type name: str
         :param definition_dict: The definition for running a recipe of this type
         :type definition_dict: dict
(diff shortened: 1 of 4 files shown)


JohnPTobe:
I don't think a new endpoint is needed. We should just remove the warning and skip difference validation if the recipe type doesn't exist and assume it's a new recipe.  If the edit api is called next with the same name it'll handle it gracefully and the user will be warned then.

JohnPTobe:
Do we want to close this PR or update the validation method to not warn and just send an empty diff?

mikenholt:
@JohnPTobe Is this what you meant?

scale/recipe/models.py
```diff
@@ -1392,11 +1392,6 @@ def validate_recipe_type_v6(self, name, definition_dict):
                 json = convert_recipe_diff_to_v6_json(diff)
                 diff = rest_utils.strip_schema_version(json.get_dict())
 
-            except RecipeType.DoesNotExist as ex:
```
JohnPTobe: I think we still need to catch the exception, just ignore it instead of adding a warning.

Issue #1717, opened 2019-07-17
Optimize Endpoint Response Creation


emimaesmith:
**Pain Point? Please describe.**
Certain API endpoints can be slow to respond at times when returning large lists of results: Ex. filtering the results for the jobs page.

**Desired Solution**
The filtering on certain endpoints should be re-looked at to see if any optimization can be done to improve the responsiveness.

First reported in #1673

mheppner:
JT, Mike probably already let you know the issue, but I'll add some details here in case he didn't. `/jobs` and `/recipes` are experiencing the `n+1` query problem, despite `select_related()` being applied in the query manager. Specifically, `job_type` and `job_type_rev` are being selected for each result in `/jobs`, and `recipe_type` from `/recipes` (I didn't see a query for `recipe_type_rev`). Removing those fields from the serializer removes the extra queries.

mikenholt:
Mark did a very quick inspection of this last week. Here are his findings as it relates to the jobs and recipes pages (the two most important pages IMO): 

![Screenshot 2019-07-17 13 39 21](https://user-images.githubusercontent.com/22593495/61397644-5836ea00-a898-11e9-8785-507282e1a05c.png)

PR #1735, opened 2019-07-31
Issue 1717 - Optimize API responsiveness for jobs and recipes

mheppner:
##### Checklist

- [x] `manage.py test` passes

### Affected app(s)

- jobs
- recipes

### Description of change

`/v6/jobs/` and `/v6/recipes/` were both performing extra queries, `n+1` for each item returned, as outlined in #1717.

For jobs, despite being in `select_related()`, the nested `job_type_rev__job_type` was being queried for each result. The docs say that fields in the `defer()` list can also be `select_related()`, but maybe this doesn't play well with DRF serializers? Similar situation for recipes endpoint.

I didn't put this change in the manager (where the `select_related()`/`defer()` currently happens) in case this messes up performance elsewhere.

For the jobs endpoint, this cuts queries from 36 (449 ms) to 4 (152 ms). Recipes endpoint is similar, but not as extreme.

### DIFF
--- a/scale/job/views.py
+++ b/scale/job/views.py
@@ -631,6 +631,7 @@ class JobsView(ListAPIView):
                                        batch_ids=batch_ids, recipe_ids=recipe_ids,
                                        error_categories=error_categories, error_ids=error_ids,
                                        is_superseded=is_superseded, order=order)
+        jobs = jobs.select_related('job_type_rev__job_type').defer(None)
         page = self.paginate_queryset(jobs)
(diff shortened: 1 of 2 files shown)


JohnPTobe:
I have a fix for the travis build issue in my branch

mheppner:
Added some comments...are those descriptive enough?

Which branch has the Travis fixes?

JohnPTobe:
My PR for deprecation warnings

Issue #1627, opened 2019-05-13
Old Nodes in Scale


JohnPTobe:
Scale seems to be holding on to nodes no longer in mesos when scaling in/out. The code determining when to mark a node as offline needs to be debugged.

PR #1739, opened 2019-08-09
#1627 - Old Nodes in Scale

mikenholt:
##### Checklist

- [x] migrations are included
- [x] tests are included

### Affected app(s)

 - Node
- Scheduler
### Description of change

Modified the mesos offers logic to update a new Node field (last_offer_received) every time an offer for a node was received. An evaluation is done every 10 seconds where a node that is not running jobs and has not received offers in 5 minutes will be marked as `is_active=False` and purged from the node list that lives in memory for the scheduler.

Removed one test as I don't see why a node would be marked as `is_active=False` and we would want that to persist after a node restarts.  If there is a reason why we need that let me know.

I didn't add `last_offer_received` to the serializer, but I can if desired. I see as an internal-to-Scale field only.


mikenholt:
I'll fix the test when I get a chance.

scale/job/models.py
```diff
@@ -1271,11 +1271,11 @@ def get_job_exes(self, job_id, started=None, ended=None, statuses=None, node_ids
             if 'RUNNING' in statuses:
                 # This is a special case where we have to use exclusion so that running executions (no job_exe_end) are
                 # included
-                exclude_statues = []
+                exclude_statuses = []
```
JohnPTobe: take those statues down!

scale/scale/local_settings2.py
```diff
@@ -0,0 +1,53 @@
...
+# Zookeeper URL for scheduler leader election. If this is None, only a single not is used and election isn't performed.
+SCHEDULER_ZK = None
+
+# The full name for the Scale Docker image (without version tag)
+SCALE_DOCKER_IMAGE = 'geoint/scale'
+BROKER_URL = 'amqp://guest:guest@localhost:55672//'
+
+POSTGIS_TEMPLATE = 'template_postgis'
+
+DATABASES = {'default': dj_database_url.config(default='postgis://scale:scale@localhost:55432/scale')}
```
JohnPTobe: accidentally added while testing?

mikenholt: Yeah this doesn't belong.....

cshamis:
I like it!

Random emoji!

🎳

PR #1749, opened 2019-08-21
Batches

emimaesmith:
##### Checklist

- [ ] `manage.py test` passes
- [x] tests are included
- [ ] migrations are included

### Affected app(s)

batch
data

### Description of change

Updated the batches API to allow for creation of a batch without having a previous batch.


scale/batch/messages/create_batch_recipes.py
```diff
@@ -66,20 +74,24 @@ def from_json(json_dict):
...
+        # Reprocess recipes from previous batch or create a new batch
         if not self.is_prev_batch_done:
-            new_messages.extend(self._handle_previous_batch(batch, definition))
+            if batch.superseded_batch_id:
+                new_messages.extend(self._handle_previous_batch(batch, definition))
+            else:
+                new_messages.extend(self._handle_new_batch(batch, definition))
 
         if self.is_prev_batch_done:
             logger.info('All re-processing messages created, marking recipe creation as done')
```
JohnPTobe: add different log message for new batch vs reprocess

scale/batch/messages/create_batch_recipes.py
```diff
@@ -91,6 +103,80 @@ def execute(self):
...
+        messages = []
+
+        dataset = DataSet.objects.get(pk=definition.dataset)
+        dataset_definition = dataset.get_definition()
+        recipe_type = RecipeType.objects.get(pk=batch.recipe_type_id)
+        recipe_inputs = recipe_type.get_definition().get_input_keys()
+        
+        # No recipe inputs match the dataset 
+        if not any(elem in recipe_inputs for elem in dataset_definition.param_names):
+            return messages
```
JohnPTobe: log error here? we should never have a batch defined where dataset parameters don't match the recipe inputs

scale/batch/definition/definition.py
```diff
@@ -48,6 +49,8 @@ def validate(self, batch):
                 self.prev_batch_diff.set_force_reprocess(self.forced_nodes)
             if not self.prev_batch_diff.can_be_reprocessed:
                 raise InvalidDefinition('PREV_BATCH_NO_REPROCESS', 'Previous batch cannot be reprocessed')
+        # New batch
```
JohnPTobe: validate dataset parameters against recipe revision inputs?

scale/batch/messages/create_batch_recipes.py
```diff
@@ -91,6 +103,80 @@ def execute(self):
...
+
+        # Get previous recipes for dataset files:
+        ds_files = DataSetFile.objects.get_dataset_files(dataset.id).values_list('scale_file_id', flat=True)
+        recipe_ids = RecipeInputFile.objects.filter(input_file_id__in=ds_files).values_list('recipe_id', flat=True)
```
JohnPTobe: Not sure about reprocessing with a new batch. I think we only want to reprocess when superseding an old batch.  If we are reprocessing we need to also filter by recipe type and revision so we reprocess the correct recipe.

scale/batch/models.py
```diff
@@ -100,6 +100,77 @@ def create_batch_v6(self, title, description, recipe_type, event, definition, co
...
+        #      - If the dataset has a global parameter matching the input of the 
+        #        recipe type, count the number of files in each dataset member
+        #      - If the dataset has a parameter matching the input of the recipe
+        #        type, count the number of files in each member that matches the parameter
+        
+        from recipe.models import RecipeType
+        from data.models import DataSet, DataSetMember, DataSetFile
+        dataset = DataSet.objects.get(pk=definition.dataset)
+        dataset_definition = dataset.get_definition()
+        recipe_type = RecipeType.objects.get(pk=batch.recipe_type_id)
```
JohnPTobe: use recipe type revision here as the definition could have changed since the batch was created

scale/batch/models.py
```diff
@@ -100,6 +100,77 @@ def create_batch_v6(self, title, description, recipe_type, event, definition, co
...
+                    estimated_recipes += (1 + RecipeTypeSubLink.objects.count_subrecipes(sub_type_id, recurse=True)) * num_files
+                
+                # If it's a child of a forced job node, we're going to need to run it
+                else:
+                    recipe_type_def = recipe_type.get_definition()
+                    for job_node in definition.forced_nodes.get_forced_node_names():
+                        if recipe_type_def.has_descendant(job_node, sub):
+                            estimated_recipes += (1 + RecipeTypeSubLink.objects.count_subrecipes(sub_type_id, recurse=True)) * num_files
+            
+        # if not (definition.forced_nodes and definition.forced_nodes.all_nodes):
```
JohnPTobe: delete commented out code

scale/batch/test/test_views.py
```diff
@@ -211,10 +216,264 @@ def test_create_invalid_configuration(self):
...
+        result = json.loads(response.content)
+        new_batch_id = result['id']
+        self.assertIsNotNone(new_batch_id)
+        
+        # Verify all recipes (including sub-recipes) were created:
+        # 2 base recipes, each with two sub-recipes = 6 recipes
+        self.assertEqual(result['recipes_estimated'], 6)
+        mock_create.assert_called_with(new_batch_id)
+        
+        # import pdb; pdb.set_trace()
```
JohnPTobe: Is this test not working or no longer needed?

scale/batch/test/test_views.py
```diff
@@ -860,3 +1119,63 @@ def test_create_successful(self):
...
+            }
+        }
+        
+        url = '/v6/batches/validation/'
+        response = self.client.generic('POST', url, json.dumps(json_data), 'application/json')
+        self.assertEqual(response.status_code, status.HTTP_200_OK, response.content)
+        result = json.loads(response.content)
+        self.assertTrue(result['is_valid'])
+        self.assertEqual(result['recipes_estimated'], 4) # two files, each with one sub-recipe
+
```
JohnPTobe: If we can easily add a test somewhere that checks the created jobs' priority that would be great.  If not, that can be a separate issue.

scale/data/models.py
```diff
@@ -379,7 +379,7 @@ def create_dataset_members(self, dataset, data_list):
                 dataset_member.data = convert_data_to_v6_json(d).get_dict()
                 dataset_member.file_ids = list(data_util.get_file_ids(d))
                 dataset_members.append(dataset_member)
-                datasetfiles = DataSetFile.objects.create_dataset_files(dataset, d, existing_scale_ids)
```
JohnPTobe: d'oh! good catch

scale/recipe/definition/definition.py
```diff
@@ -418,3 +418,37 @@ def _topological_order_visit(self, node, results, perm_set, temp_set):
...
+        """Recursive function for checking children
+        
+        """
+        for child in parent_node.children.values():
+            if child.name == descendant:
+                return True
+            
+            if child.node_type != RecipeNodeDefinition.NODE_TYPE:
+                return self._has_descendant(child, descendant)
+
```
JohnPTobe: Don't we need to return False here?

scale/batch/test/test_views.py
emimaesmith: Added a portion to the `TestBatchesViewV6.test_create_priority` that tests the new batch with an updated priority.

scale/batch/definition/definition.py
```diff
@@ -49,8 +50,19 @@ def validate(self, batch):
...
                 raise InvalidDefinition('PREV_BATCH_NO_REPROCESS', 'Previous batch cannot be reprocessed')
-        # New batch
         
+        # New batch - need to validate dataset parameters against recipe revision
+        elif self.dataset:
+            from data.models import DataSet
+            from recipe.models import RecipeType
+            dataset = DataSet.objects.get(pk=self.dataset)
+            dataset_definition = dataset.get_definition()
+            recipe_type = RecipeType.objects.get(name=batch.recipe_type.name, revision_num=batch.recipe_type_rev.revision_num)
```
JohnPTobe: Use RecipeTypeRevision.objects.get_revision() method and get the definition from the revision. The recipetype get method will fail here if the revision number requested isn't the latest version.  The RecipeTypeRevision class stores the definition for the revision and the RecipeType class stores the elements that are common to each revision.

scale/batch/definition/definition.py
```diff
@@ -49,8 +50,19 @@ def validate(self, batch):
...
+            recipe_inputs = recipe_type.get_definition().get_input_keys()
+
+            # No recipe inputs match the dataset 
+            if not any(elem in recipe_inputs for elem in dataset_definition.param_names):
+                raise InvalidDefinition('NO_MATCHING_PARAMS', 'No parameters in the dataset match the recipe type inputs')
```
JohnPTobe: Good, but we need to make sure the dataset has all of the required parameters for the recipe.  You can do this by calling recipe_definition.input_interface.validate_connection(dataset_definition.parameters), although then you'll be missing the global parameters so you'd need to combine the two interface objects and then call the validate_connection method.

scale/batch/messages/create_batch_recipes.py
```diff
@@ -116,41 +119,45 @@ def _handle_new_batch(self, batch, definition):
         """
         
         messages = []
-
         dataset = DataSet.objects.get(pk=definition.dataset)
         dataset_definition = dataset.get_definition()
-        recipe_type = RecipeType.objects.get(pk=batch.recipe_type_id)
+        recipe_type = RecipeType.objects.get(name=batch.recipe_type.name, revision_num=batch.recipe_type_rev.revision_num)
```
JohnPTobe: again, RecipeTypeRevision

scale/batch/messages/create_batch_recipes.py
```diff
@@ -116,41 +119,45 @@ def _handle_new_batch(self, batch, definition):
...
         recipe_inputs = recipe_type.get_definition().get_input_keys()
         
         # No recipe inputs match the dataset 
         if not any(elem in recipe_inputs for elem in dataset_definition.param_names):
+            logger.info('None of the dataset parameters matched the recipe type inputs; No recipes will be created')
```
JohnPTobe: Same validation_connection as before

scale/batch/messages/create_batch_recipes.py
```diff
@@ -48,7 +48,7 @@ def __init__(self):
         super(CreateBatchRecipes, self).__init__('create_batch_recipes')
 
         self.batch_id = None
-        self.is_prev_batch_done = False  # Indicates if all recipes from pervious batch have been handled
```
JohnPTobe: 👀

cshamis:
Brilliant!

Issue #1605, opened 2019-04-30
Improve metrics to show live stats


cshamis:
Daily summaries or hourly are good, but for dashboard the UI needs to have "live" statistics.  Waiting an hour to update on ingests or running counts is unacceptable.

cshamis:
Submitted as a backend ticket becuase UI already expects live data.

PR #1761, opened 2019-08-31
Revert "#1605 Bin Metrics By Hour"

emimaesmith:
Reverts ngageoint/scale#1732 
Once we release 7.0.1 and are able to update all the charts on the UI we can put this change back in.


cshamis:
So. .. We get things stable and we shift the instrumentation (dashboard) changes to 7.1 with forking?

Issue #1768, opened 2019-09-04
Transfer_suffix is required for directory strikes


Fizz11:
**Description**
Transfer_suffix is currently optional in the UI, however the strike will fail if it is missing. 

**Reproduction Steps**
Steps to reproduce the problem:
1. make a directory strike

**Additional context**
easy fix is to set default value to _tmp if value is not provided in the UI. also update documentation to label it optional.

emimaesmith:
Closed via #1771

PR #1771, opened 2019-09-04
#1768 - Transfer_suffix required for directory strikes

Fizz11:
##### Checklist

- [x] `manage.py test` passes
- [x] tests are included

### Affected app(s)

ingest

### Description of change

made transfer_suffix optional for directory strikes.

### DIFF
--- a/scale/docs/rest/v6/strike.rst
+++ b/scale/docs/rest/v6/strike.rst
@@ -523,7 +523,7 @@ A strike configuration JSON describes a set of configuration settings that affec
 +----------------------------+----------------+----------+--------------------------------------------------------------------+
 | .transfer_suffix           | String         | Optional | (dir-watcher)Defines a suffix that is used on the file names to    |
 |                            |                |          | indicate that files are still transferring and have not yet        |
-|                            |                |          | finished being copied into the monitored directory                 |
+|                            |                |          | finished being copied. Defaults to '_tmp'                          |
 +----------------------------+----------------+----------+--------------------------------------------------------------------+
 | .sqs_name                  | String         | Required | (s3) Name of the SQS queue that should be polled for object        |
 |                            |                |          | creation notifications that describe new files in the S3 bucket.   |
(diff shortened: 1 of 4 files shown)


cshamis:
Transfer suffix may be optional, but without it, strike will have NO WAY of knowing if a file is done transferring or not.

The transfer suffix should only be left undefined if the user KNOWS FOR CERTAIN that transfers are atomic.

We might want to warn them of that.

scale/docs/rest/v6/strike.rst
```diff
@@ -521,7 +521,7 @@ A strike configuration JSON describes a set of configuration settings that affec
 +----------------------------+----------------+----------+--------------------------------------------------------------------+
 | .type                      | String         | Required | The type of the monitor. Must be either 'dir-watcher' or 's3'      |
 +----------------------------+----------------+----------+--------------------------------------------------------------------+
-| .transfer_suffix           | String         | Required | (dir-watcher)Defines a suffix that is used on the file names to    |
+| .transfer_suffix           | String         | Optional | (dir-watcher)Defines a suffix that is used on the file names to    |
```
emimaesmith: I would document here that if it's not provided, then a default of '_tmp' will be used.

scale/ingest/strike/monitors/dir_monitor.py
```diff
@@ -78,13 +81,11 @@ def stop(self):
...
         """See :meth:`ingest.strike.monitors.monitor.Monitor.validate_configuration`
         """
-
-        if 'transfer_suffix' not in configuration:
-            raise InvalidMonitorConfiguration('transfer_suffix is required for dir-watcher monitor')
-        if not isinstance(configuration['transfer_suffix'], basestring):
-            raise InvalidMonitorConfiguration('transfer_suffix must be a string')
-        if not configuration['transfer_suffix']:
-            raise InvalidMonitorConfiguration('transfer_suffix must be a non-empty string')
+        if 'transfer_suffix' in configuration:
```
emimaesmith: Is there any way to return a warning if transfer_suffix isn't defined? Something that will return saying something like "No transfer suffix defined, '_tmp' will be used"?

scale/ingest/test/strike/monitors/test_dir_monitor.py
```diff
@@ -18,7 +18,7 @@ def test_validate_configuration_missing_transfer_suffix(self):
         config = {
             'type': 'dir-watcher'
         }
-        self.assertRaises(InvalidMonitorConfiguration, DirWatcherMonitor().validate_configuration, config)
+        DirWatcherMonitor().validate_configuration(config)
```
emimaesmith: If we don't care about transfer_suffix being there or not, might as well just remove this test.

Issue #1775, opened 2019-09-05
Input metadata manifest json


cshamis:
Does not formulate the permalink URL from the workspace url prefix.  Just gives the user the workspace ID.  

So, that kinda stinks.

JohnPTobe:
I added some code to test this and from what I can tell this shouldn't have been working at all because of a bug in determining if the seed manifest has INPUT_METADATA_MANIFEST as an input.

JohnPTobe:
After fixing that bug, here is what I get for the json:
`

{
	"JOB": {
		"input_2": [{
			"id": 1,
			"file_name": "my_test_file.txt",
			"workspace": {
				"id": 22,
				"name": "test-workspace-22"
			},
			"data_type_tags": [],
			"media_type": "text/plain",
			"file_type": "SOURCE",
			"file_size": 100,
			"file_path": "path/1/file1.txt",
			"is_deleted": false,
			"url": "http://test.com/path/1/file1.txt",
			"created": "2019-09-06T14:46:24.891612Z",
			"deleted": null,
			"data_started": null,
			"data_ended": null,
			"source_started": null,
			"source_ended": null,
			"source_sensor_class": null,
			"source_sensor": null,
			"source_collection": null,
			"source_task": null,
			"last_modified": "2019-09-06T14:46:24.891636Z",
			"geometry": null,
			"center_point": null,
			"countries": [],
			"job_type": null,
			"job": null,
			"job_exe": null,
			"job_output": null,
			"recipe_type": null,
			"recipe": null,
			"recipe_node": null,
			"batch": null,
			"is_superseded": false,
			"superseded": null,
			"meta_data": {}
		}],
		"input_3": [{
			"id": 2,
			"file_name": "my_test_file.txt",
			"workspace": {
				"id": 22,
				"name": "test-workspace-22"
			},
			"data_type_tags": [],
			"media_type": "text/plain",
			"file_type": "SOURCE",
			"file_size": 100,
			"file_path": "path/2/file2.txt",
			"is_deleted": false,
			"url": "http://test.com/path/2/file2.txt",
			"created": "2019-09-06T14:46:24.901146Z",
			"deleted": null,
			"data_started": null,
			"data_ended": null,
			"source_started": null,
			"source_ended": null,
			"source_sensor_class": null,
			"source_sensor": null,
			"source_collection": null,
			"source_task": null,
			"last_modified": "2019-09-06T14:46:24.901175Z",
			"geometry": null,
			"center_point": null,
[...]

PR #1776, opened 2019-09-06
#1775 - Input metadata manifest json

JohnPTobe:
##### Checklist

- [x] `manage.py test` passes
- [x] tests are included

### Affected app(s)
- job

### Description of change
Fixing input metadata manifest so clint can actually write new publishers.

### DIFF
--- a/scale/job/data/job_data.py
+++ b/scale/job/data/job_data.py
@@ -328,7 +328,7 @@ class JobData(object):
                 env_var_name = normalize_env_var_name(file_input.name)
                 if file_input.name not in interface.parameters:
                     logger.warning("File input %s not specified in interface %s" % (file_input.name, interface))
-                if interface.parameters[file_input.name].multiple:
+                if file_input.name in interface.parameters and interface.parameters[file_input.name].multiple:
                     # When we have input for multiple files, map in the entire directory
                     env_vars[env_var_name] = os.path.join(SCALE_JOB_EXE_INPUT_PATH, file_input.name)
                 else:
--- a/scale/job/management/commands/scale_pre_steps.py
+++ b/scale/job/management/commands/scale_pre_steps.py
@@ -123,7 +123,8 @@ class Command(BaseCommand):
                                                    input_data.values[i].file_ids]
 
         logger.debug("Scale Input Metadata Manifest Generated:")
-        logger.debug(input_metadata)
+        log_str = json.dumps(input_metadata, sort_keys=True, indent=4, separators=(',', ': '))
+        logger.debug(log_str)
 
         try:
             with open(SCALE_INPUT_METADATA_PATH, 'w') as metadata_file:
--- a/scale/job/management/commands/scale_pre_steps.py
+++ b/scale/job/management/commands/scale_pre_steps.py
@@ -107,6 +107,7 @@ class Command(BaseCommand):
         # Generate input metadata dict
         input_metadata = {}
         config = job_exe.get_execution_configuration()
+        print config.get_dict()
         if 'input_files' in config.get_dict():
             input_metadata['JOB'] = {}
             input_data = job_exe.job.get_input_data()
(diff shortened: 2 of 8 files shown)


scale/job/data/job_data.py
```diff
@@ -328,7 +328,7 @@ def get_injected_env_vars(self, input_files, interface):
                 env_var_name = normalize_env_var_name(file_input.name)
                 if file_input.name not in interface.parameters:
```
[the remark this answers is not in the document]
JohnPTobe: I ran into this bug while testing.  I think it's good as is. It's perfectly fine to have extra files that aren't specified in the interface and we need to handle that so no need for an error.  Also, I believe normally data should be sanitized elsewhere to throw out extra parameters so I don't think this is an issue in practice, I just added this so my tests would work without too much more effort.

JohnPTobe:
Yes, I added tests to ensure the metadata is produced.  The public url is present as long as the workspace has a base url.  I put a sample generated json file made with the test I added in the issue.

scale/job/management/commands/scale_pre_steps.py
```diff
@@ -126,8 +126,12 @@ def _generate_input_metadata(self, job_exe):
                     input_metadata['RECIPE'][i] = [serialize(ScaleFile.objects.get_details(file_id=f)).data for f in
                                                    input_data.values[i].file_ids]
 
-        workspace_names = config.get_input_workspace_names()
-        workspace_models = {w.name: w for w in Workspace.objects.get_workspaces(names=workspace_names)}
+        name = job_exe.job.get_job_configuration().get_output_workspace('input_metadata_manifest')
```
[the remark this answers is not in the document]
gisjedi: Based on our discussion we can remove the use of output workspace.

scale/job/management/commands/scale_pre_steps.py
```diff
@@ -143,17 +147,16 @@ def _generate_input_metadata(self, job_exe):
...
-                        if not input_metadata_id:
-                            ScaleFile.objects.upload_files(workspace, [FileUpload(scale_file, local_path)])
-                            input_metadata_id = ScaleFile.objects.get(file_name=file_name).id
-                            data = job_exe.job.get_job_data()
-                            data.add_file_input('INPUT_METADATA_MANIFEST', input_metadata_id)
-                            job_exe.job.input = data.get_dict()
-                            job_exe.job.save()
-                    except:
-                        continue
+                try:
```
gisjedi: Should be able to remove this entire block as uploading the metadata won't be necessary.

scale/job/seed/manifest.py
```diff
@@ -417,7 +417,10 @@ def needs_input_metadata(self):
         :rtype: bool
         """
 
-        return 'INPUT_METADATA_MANIFEST' in self.get_input_files()
```
gisjedi: We can entirely remove the needs_input_metadata check against the manifest. May still want evaluation for if no input at all exists. As long is it can create empty metadata json, we are good.

JohnPTobe: done

JohnPTobe:
Ok, I'm happy with these changes. Mike can proceed with testing these and once his test passes we should be good.

scale/job/management/commands/scale_pre_steps.py
```diff
@@ -126,36 +122,14 @@ def _generate_input_metadata(self, job_exe):
                     input_metadata['RECIPE'][i] = [serialize(ScaleFile.objects.get_details(file_id=f)).data for f in
                                                    input_data.values[i].file_ids]
 
-        workspace_names = config.get_input_workspace_names()
-        workspace_models = {w.name: w for w in Workspace.objects.get_workspaces(names=workspace_names)}
+        logger.debug("Scale Input Metadata Manifest Generated:")
+        logger.debug(input_metadata)
```
gisjedi: Minor formatting concern. This will output the dict, not json and you'll have all the unicode prefix mess in the output. Can you do something like here (https://github.com/ngageoint/scale/blob/e33d510d066ae8a033a863e983bdb1710b868336/scale/recipe/messages/process_condition.py#L94) to improve our sanity?

PR #1779, opened 2019-09-10
Mesoshttp 0.3.2 for 5.9.6

gisjedi:
This PR includes all changes that will be applied to the 5.9.6 release of Scale. This is to resolve the issue with Scale losing connection to Mesos and failing to re-establish itself as a framework during an extended network partition.


gisjedi:
This has been squashed down for a 5.9.6 tag in https://github.com/ngageoint/scale/tree/5.9.6-prerelease. I've tested for proper scheduling behavior in AIS Omega cluster and am cutting the release based on instructions here: https://github.com/ngageoint/scale/wiki/Creating-a-patch-release

PR #1781, opened 2019-09-10
Input metadata small fix

mikenholt:
- The `SCALE_INPUT_METADATA_PATH` file doesn't exist so it needs the `+` operator on open.
- The `tmp` folder does not exist so it needs to be made.


cshamis:
Why the tmp folder?  That seems crappy.

mikenholt:
> Why the tmp folder? That seems crappy.
> […](#)

fixed

mikenholt:
> Are you sure the mkdir won't cause an exception in the pre_step (not in the unit test, but in a deployed state) when it already exists? Python is real picky about this: https://stackoverflow.com/questions/600268/mkdir-p-functionality-in-python

GREAT CALL.  That line was needed when `/tmp` was in the path but I think it is obsolete now that the metadata file lives in the root folder for inputs.

Issue #1785, opened 2019-09-13
Allow auto_update on job type patch


JohnPTobe:
**Description**
Currently auto_update isn't an allowed field when patching a job type.  Also, auto_update should default to True wherever it is present. Finally, invalid field errors should be caught and return a useful error message instead of a 500 error.

**Reproduction Steps**
Steps to reproduce the problem:
1. Make job type patch call to API with auto_update field present
2. Receive 500 error

PR #1787, opened 2019-09-15
#1785 : Allow auto_update on job type patch

JohnPTobe:
##### Checklist

- [x] `manage.py test` passes

### Affected app(s)
- job

### Description of change
Fixed 500 errors for invalid job type fields, added auto_update as a valid field for job type patch endpoint and made auto_update default to true. The last one doesn't help when using the UI so the UI needs to be updated to default to true as well.

### DIFF
--- a/scale/job/views.py
+++ b/scale/job/views.py
@@ -111,7 +111,7 @@ class JobTypesView(ListCreateAPIView):
         manifest_dict = rest_util.parse_dict(request, 'manifest', required=True)
 
         # If editing an existing job type, automatically update recipes containing said job type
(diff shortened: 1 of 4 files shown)


scale/job/views.py
```diff
@@ -144,7 +144,7 @@ def create_v6(self, request):
                   'auto_update', 'is_active', 'is_paused'}
         for key, value in request.data.iteritems():
             if key not in fields:
-                raise InvalidJobField
+                raise BadParameter('%s is not a valid field. Valid fields are: %s' % (key, fields))
```
gisjedi: This is a fantastic improvement to know what caused the failure.

Issue #1778, opened 2019-09-10
Empty Mesos Resources Response


JohnPTobe:
**Description**
Mesos can get messed up after restarting masters and scale gets an empty response for get_agent_resources.  We need to catch the value error in functions calling get_agent_resources (scale/mesos_api/unversioned/agent.py) and handle it. By either retrying, raising an error up to the operator that scale needs to be restarted or something else.

**Reproduction Steps**
Steps to reproduce the problem:
1. Run scale 
2. Wait for mesos to hiccup
3. ???
4. Nothing runs in scale

**Expected behavior**
Scale keeps running or tells users there is a problem that needs to be fixed

PR #1795, opened 2019-09-25
#1778 - Empty Mesos Resources Response

JohnPTobe:
##### Checklist

- [x] `manage.py test` passes
- [x] tests are included

### Affected app(s)
- scheduler

### Description of change
Catch exceptions encountered when getting agent resources from mesos (caused by masters restarting or anything else) and shutdown scale if exceptions continue to occur for 30 minutes. We may want to pause the scheduler and/or raise an error instead of shutting down.

### DIFF
--- a/scale/scheduler/resources/manager.py
+++ b/scale/scheduler/resources/manager.py
@@ -32,6 +32,7 @@ class ResourceManager(object):
         self._new_offers = {}  # {Offer ID: ResourceOffer}
         self._new_offers_lock = threading.Lock()  # Protects self._new_offers
         self._mesos_error = None
+        self._mesos_error_started = None
 
     def add_new_offers(self, offers):
         """Adds new resource offers to the manager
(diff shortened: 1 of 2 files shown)


scale/scheduler/resources/manager.py
```diff
@@ -285,8 +290,25 @@ def sync_with_mesos(self, host_address):
...
+            if not self._mesos_error:
+                self._mesos_error_started = datetime.now()
+        except Exception as ex:
+            logger.exception('Error getting agent resource totals from Mesos: %s' % ex)
+            self._mesos_error = ex.message
+            if not self._mesos_error:
+                self._mesos_error_started = datetime.now()
+
+        if self._mesos_error_started and datetime.now() > self._mesos_error_started + SHUTDOWN_PERIOD:
+            from scheduler.management.commands.scale_scheduler import GLOBAL_SHUTDOWN
```
gisjedi: Lets log either here or in GLOBAL_SHUTDOWN (_shutdown in scale_scheduler) that we are forcibly going down.

scale/scheduler/resources/manager.py
```diff
@@ -285,8 +290,25 @@ def sync_with_mesos(self, host_address):
...
+            GLOBAL_SHUTDOWN()
+
+        if resources:
+            #clear error
```
gisjedi: Log here (INFO) that we are receiving resources again.

scale/scheduler/resources/manager.py
gisjedi: After thinking about it lets log here the reason for the shutdown.

JohnPTobe:
I tested on the high side but scale shutdown before my code could run due to zookeeper session being lost.

Issue #1798, opened 2019-10-02
return priority with job details in API


Fizz11:
**Description**
job_type priority is not displayed in the UI when viewing job details. this is because priority is not provided with the job details API.

**Reproduction Steps**
Navigate to Processing -> Jobs
Select a job in the list
Note the Priority field is blank

**Expected behavior**
priority field not be blank

emimaesmith:
Closed via #1799

PR #1799, opened 2019-10-02
#1798 - return priority with job details in API

Fizz11:
##### Checklist

- [x] `manage.py test` passes
- [x] tests are included

### Affected app(s)
jobs

### Description of change
added configuration return to job details endpoint

### DIFF
--- a/scale/job/models.py
+++ b/scale/job/models.py
@@ -391,7 +391,12 @@ class JobManager(models.Manager):
                 job.execution = None
         else:
             job.execution = None
-        job.configuration =  convert_config_to_v6_json(job.get_job_configuration()).get_dict()
+
+        configuration = job.get_job_configuration()
+        manifest = SeedManifest(job.job_type.manifest, do_validate=False)
+        configuration.remove_secret_settings(manifest)
+        job.configuration = convert_config_to_v6_json(configuration).get_dict()
+        
         # Attempt to get related recipe
         # Use a localized import to make higher level application dependencies optional
         try:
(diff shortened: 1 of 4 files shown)


scale/job/models.py
```diff
@@ -391,7 +391,7 @@ def get_details(self, job_id):
                 job.execution = None
         else:
             job.execution = None
-
+        job.configuration = job.job_type.configuration
```
JohnPTobe: We should use the get_configuration method to scrub secrets or something similar

scale/job/models.py
```diff
@@ -391,7 +391,7 @@ def get_details(self, job_id):
                 job.execution = None
         else:
             job.execution = None
-        job.configuration = job.job_type.configuration
+        job.configuration =  convert_config_to_v6_json(job.get_job_configuration()).get_dict()
```
JohnPTobe: Sorry, I wasn't clear earlier because I forgot exactly how I did it. You need to scrub the secrets like we're doing in the job type get_details method.

Fizz11: think i got it this time. should be OK as long as we call remove_secret_settings correct?

scale/job/models.py
```diff
@@ -392,6 +392,11 @@ def get_details(self, job_id):
         else:
             job.execution = None
 
+        configuration = job.get_job_configuration()
+        manifest = SeedManifest(job.job_type.manifest, do_validate=False)
```
JohnPTobe: almost there. we should get the manifest for the job type revision, not necessarily the latest one. Probably use the get_job_interface method.

JohnPTobe: Also we should add a test that DB_PASS is removed

PR #1807, opened 2019-10-15
1806 - Scale Doesn't Handle Optional Output->Inputs

Fizz11:
##### Checklist

- [x] `manage.py test` passes
- [ ] tests are included

### Affected app(s)
job

### Description of change
when an input is missing, check to see if it was optional in the first place before throwing an exception.

### DIFF
--- a/scale/recipe/definition/node.py
+++ b/scale/recipe/definition/node.py
@@ -9,7 +9,9 @@ from data.data.data import Data
 from data.interface.exceptions import InvalidInterfaceConnection
 from data.interface.interface import Interface
 from recipe.definition.exceptions import InvalidDefinition
-
+from data.data.exceptions import InvalidData
+import logging
+logger = logging.getLogger(__name__)
 
 class NodeDefinition(object):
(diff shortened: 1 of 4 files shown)


Fizz11:
currently creating a new test for this specific case

scale/recipe/definition/node.py
```diff
@@ -80,9 +82,16 @@ def generate_input_data(self, recipe_input_data, node_outputs):
...
         for connection in self.connections.values():
-            connection.add_value_to_data(input_data, recipe_input_data, node_outputs)
+            try:
+                connection.add_value_to_data(input_data, recipe_input_data, node_outputs)
+            except InvalidData as ex:
+                if not connection.output_name in optional_outputs:
+                    logger.warning("output name %s not found in optional_outputs", connection.output_name)
+                    raise
+                else:
+                    logger.info("InvalidData exception occured due to optional output not present. proceeding with job execution.")
```
emimaesmith: Capitalize the `p` or change the `.` to a `;`.

Issue #1808, opened 2019-10-16
Recipe Dependency Checking


emimaesmith:
**Description**
The API is currently validating recipe types that contain sub-recipes of itself.  

**Reproduction Steps**
Steps to reproduce the problem:
1. Create an empty recipe type (Recipe type with no job/recipe/condition nodes)
2. Edit the empty recipe type.
3. Open the Recipe Nodes picker and select the empty recipe
4. Click Validate - Note it succeeds
5. Click Save - Note it succeeds

**Expected behavior**
The validation should fail when the recipe type detects itself as a sub-recipe.

**Screenshots**
![image](https://user-images.githubusercontent.com/29577288/66935626-505b2300-f00a-11e9-90d7-02eca1413ec6.png)

**Version and Environment Details**
 - Scale API: 7.0.1
 - Scale UI: 0.9.0

**Additional context**
I believe the cyclic dependencies are checking both recipe type name and recipe type revision to determine dependency. This check should fail if a sub-recipe of the same recipe type name as the parent is detected. It would also be ideal if the UI would hide the recipe type being edited from the recipe node selection.

PR #1809, opened 2019-10-17
Validate Recipes Don't Contain Sub-recipes of Themselves

emimaesmith:
##### Checklist

- [x] `manage.py test` passes
- [x] tests are included

### Affected app(s)

recipe
### Description of change

Updated the Edit/Validate logic to check for sub-recipes of the same time of the recipe being edited. This validation will drill down into each sub-recipe to ensure we don't have a loop anywhere.
Addresses #1808

### DIFF
--- a/scale/recipe/models.py
+++ b/scale/recipe/models.py
@@ -1458,7 +1458,6 @@ class RecipeTypeManager(models.Manager):
         if not recipe_type_name:
             return warnings
 
-        import pdb; pdb.set_trace()
         # Verify we don't have recursive sub-recipe dependencies - need to go through all sub-recipes
         for node_name in definition.get_topological_order():
             node = definition.graph[node_name]
--- a/scale/recipe/models.py
+++ b/scale/recipe/models.py
@@ -1401,7 +1401,6 @@ class RecipeTypeManager(models.Manager):
         diff = {}
 
         definition = None
-
         try:
             definition = RecipeDefinitionV6(definition=definition_dict, do_validate=True).get_definition()
         except InvalidDefinition as ex:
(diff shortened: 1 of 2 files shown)


scale/recipe/models.py
```diff
@@ -1281,6 +1281,16 @@ def edit_recipe_type_v6(self, recipe_type_id, title=None, description=None, defi
                 recipe_type.definition = convert_recipe_definition_to_v6_json(definition).get_dict()
             else:
                 raise InvalidDefinition('INVALID_DEFINITION', 'This version of the recipe definition is invalid to save')
+
+            # try:
```
gisjedi: Meant to leave this commented out? If so, lets get rid of it.

scale/recipe/models.py
```diff
@@ -1442,6 +1451,37 @@ def validate_recipe_type_v6(self, name, definition_dict):
 
         return RecipeTypeValidation(is_valid, errors, warnings, diff)
 
+    def validate_recursive_subs(self, recipe_type_name, definition, source_recipe_type_name=None):
+        """Validates the sub-recipes to ensure we don'' end up in a recipeception loop
+        :param recipe_type_name: The name of the recipe type we're looking for
+        :type recipe_type_name: String
+        :param definition: The RecipeDefinition we're searching
+        :type definition: RecipeDefinition
```
gisjedi: Should do a full qualified python import so you can click through to definition in IDEs. Like so: https://github.com/ngageoint/scale/blob/master/scale/job/seed/manifest.py#L91

scale/recipe/models.py
```diff
@@ -1442,6 +1451,37 @@ def validate_recipe_type_v6(self, name, definition_dict):
 
         return RecipeTypeValidation(is_valid, errors, warnings, diff)
 
+    def validate_recursive_subs(self, recipe_type_name, definition, source_recipe_type_name=None):
+        """Validates the sub-recipes to ensure we don'' end up in a recipeception loop
```
gisjedi: don'' -> don't

scale/recipe/models.py
emimaesmith: fixed

emimaesmith:
@gisjedi Ready for re-review

Issue #1289, opened 2018-08-23
Parse input file meta-data


dfaller:
Parse meta-data side car files for input files if provided in the output directory in a directory/file name structure that matches input file(s). Only allow the setting of meta-data ONCE on a Scale file (log a warning). 

This replaces the old "parse results" capability in Scale. The current "parse results" capability still needs to remain intact.

Also fix the special `url` meta-data field that Scale sets on a product's or source file's meta-data. Right now the `url` appears to always be set to `null`.

gisjedi:
It wasn't really anymore work to use the sidecar concept and didn't introduce another one-off. Adding data on source files can be done for an `INPUT_FILE` on the interface by writing to `$OUTPUT_DIR/INPUT_FILE.metadata.json`.

PR #1810, opened 2019-10-17
#1289: Parse results / Source Metadata Augmentation

gisjedi:
##### Checklist

- [x] `manage.py test` passes
- [x] tests are included

### Affected app(s)
* job
* storage

### Description of change
Add the ability to capture supplemental metadata to attach to `SOURCE` files. The only reason for this is to support augmentation of metadata associated with files that domain specific jobs can provide - the Scale Ingest doesn't have the smarts for it, nor should it.

We are expanding on the use of the metadata sidecar file for input files. There are some small differences. Since we can't guarantee writable storage within the input location, we must write them to the output directory. Details on the format and location of the metadata json can be found in the documentation here: https://github.com/ngageoint/scale/wiki/Scale-Job-Inputs-and-Outputs#source-input-metadata

The reason we are using the input file name as opposed to the same name as the input file name is because we can easily allow the algorithm developers at build time to know the location, since it is defined on the interface.

_We do not support making input data updates to multiple file inputs. We also limit updates to `SOURCE` file_types._

### DIFF
--- a/scale/source/configuration/source_data_file.py
+++ b/scale/source/configuration/source_data_file.py
@@ -1,8 +1,12 @@
 """Defines the source data file input type contained within job data"""
+import logging
+
 from job.configuration.data.data_file import AbstractDataFileParseSaver
 from source.models import SourceFile
 from storage.models import ScaleFile
 from util.parse import parse_datetime
+
+logger = logging.getLogger(__name__)
 
 
 class SourceDataFileParseSaver(AbstractDataFileParseSaver):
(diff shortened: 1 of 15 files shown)


scale/source/configuration/source_data_file.py
```diff
@@ -36,3 +40,38 @@ def save_parse_results(self, parse_results, input_file_ids):
...
+    def save_parse_results_v6(self, id_to_metadata):
+        """Saves the given parse results to the source file for the given ID. All database changes occur in an atomic
+        transaction.
+
+        :param id_to_metadata: Mapping of IDs to metadata objects/
+        :type id_to_metadata: { int: class:`job.seed.metadata.SeedMetadata` }
+        """
+
+        ids = id_to_metadata.keys()
+        logger.debug('List of IDs to update: {}'.format(ids))
```
emimaesmith: Do we need these debug statements in still?

gisjedi: I left them since they were debug only. They were useful to detect which file ids were matched for updating with metadata.

Issue #1811, opened 2019-10-21
Ingest API Filename Filtering


emimaesmith:
**Pain Point? Please describe.**
The /ingest/?filename=xxxx allows for returning the ingests matching specific filenames. However, these filenames have to exactly match the filename or nothing will be returned.

**Desired Solution**
We currently do a `.filter(file_name=file_name)` when matching records with a specific filename. Propose we change this to `.filter(file_name__in=file_name)` to allow for capture of like names.

**Alternative / Workaround**
Needed to complete https://github.com/ngageoint/scale-ui/issues/355

PR #1812, opened 2019-10-24
#1811 - Filter filename/scan name by like names instead of exact match

emimaesmith:
##### Checklist

- [x] `manage.py test` passes
- [x] tests are included

### Affected app(s)

ingest
### Description of change

Updated the /ingests?file_name=xxxx and /scans?name=xxx from `file_name__in`/`name__in` in order to allow for partial matching of file_name/name. Completes API updates required for https://github.com/ngageoint/scale-ui/issues/355


cshamis:
Double check with Clint.  Combining with a date might negate the index performance.  Might not.

PR #1814, opened 2019-10-28
1736 - Strike doesn't properly reload configurations

Fizz11:
##### Checklist

- [x] `manage.py test` passes

### Affected app(s)
ingest

### Description of change
whenever the workspace of a strike is changed, the strike job will now stop and restart to apply the new workspace.


Fizz11:
all requested changes should be in now.

Issue #1815, opened 2019-10-30
Make Scheduler queue ordering configurable via API


gisjedi:
**Pain Point? Please describe.**
When using Scale within a resource constrained cluster, it would be desirable to flip the ordering of the queue from the default First In First Out (FIFO) ordering to Last In First Out (LIFO). This would reduce the pain of configuration changes being buried behind a wall of outdated queued jobs. 

**Desired Solution**
The Scheduler already has the ability to flip between FIFO and LIFO processing, but it is hidden within the Scheduler database model. We just need to add the `queue_mode` field to the Scheduler serializer to make it configurable via the API:

https://github.com/ngageoint/scale/blob/a7043945078a0aed4de2a84b38fde1370855b41c/scale/scheduler/serializers.py#L14

Issue #1802, opened 2019-10-04
Log scheduling problems


JohnPTobe:
**Description**
Currently there's no insight as to why a job isn't being scheduled.  We should log why a job isn't scheduled when processing the queue. There should be warnings when a job isn't scheduled after several attempts.  We don't want to flood the logs but if a job has been sitting in the queue for awhile there's no way to debug the problem.

Issue #1801, opened 2019-10-04
500 Unschedulable Jobs


JohnPTobe:
**Description**
If there are 500 unschedulable jobs in Scale with high priority, anything with lower priority will not be able to be scheduled.  The scheduler grabs 500 jobs (QUEUE_LIMIT) and attempts to schedule them.  If none are able to be scheduled, it does nothing and returns, only to get those same 500 jobs again the next time it runs.

**Reproduction Steps**
Steps to reproduce the problem:
1. Create a job type that requests a gpu
2. Have a cluster where only one node has a gpu resource and have something other than scale use that gpu.  Scale should see that the gpu will theoretically be available at some point.
3. Create 500 jobs of the gpu job type
4. Create other jobs with lower priority that easily fit on nodes and some with higher priority
5. The low priority jobs will never run despite space being available on nodes

**Expected behavior**
Scale should wait to return until it has scheduled 500 jobs or skip some jobs if they have not been schedulable for awhile.  The unschedulable jobs should not count towards the limit of 500 jobs.

PR #1827, opened 2019-11-25
#1801, #1802 & #1815 - Scheduler

emimaesmith:
##### Checklist

- [x] `manage.py test` passes
- [x] tests are included

### Affected app(s)

scheduler

### Description of change

#1801: Updated the scheduler logic to pack in jobs as long as we're under the queue limit (500) instead of just grabbing the first 500 from the queue regardless of schedulability.
#1802: Added logging statements to the scheduler loop.
#1815: Added queue_mode to the serializer/patch arguments to support dynamically changing the queue order from LIFO to FIFO and vice versa.

### DIFF
--- a/scale/scheduler/scheduling/manager.py
+++ b/scale/scheduler/scheduling/manager.py
@@ -349,7 +349,7 @@ class SchedulingManager(object):
 
             # If there are no longer any available nodes, break
             if not nodes:
-                logger.warning('There are no nodes available. Waiting to schedule until we have resources...')
+                logger.warning('There are no nodes available. Waiting to schedule until there are free resources...')
                 break
 
             jt = job_type_mgr.get_job_type(queue.job_type.id)
@@ -366,6 +366,7 @@ class SchedulingManager(object):
             for resource in job_exe.required_resources.resources:
                 # skip sharedmem
                 if resource.name.lower() == 'sharedmem':
+                    logger.warning('Job type %s could not be scheduled due to required sharedmem resource', jt.name)
                     continue
                 if resource.name not in max_cluster_resources._resources:
                     # resource does not exist in cluster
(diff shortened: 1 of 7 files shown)


scale/scheduler/scheduling/manager.py
```diff
@@ -418,6 +421,9 @@ def _process_queue(self, nodes, job_types, job_type_limits, job_type_resources,
                 if job_type_id in job_type_limits:
                     job_type_limits[job_type_id] -= 1
 
+            if len(scheduled_job_executions) == QUEUE_LIMIT:
+                break
```
gisjedi: 2 questions here:

- Is it possible to pack more than `QUEUE_LIMIT` in one pass? If so, we should change to `>=`
- I'd add a logger.debug statement here to indicate that queue limit is exceeded as long as we know it won't spam 1000 of those messages a second. :-)

scale/scheduler/views.py
```diff
@@ -96,6 +96,13 @@ def patch_v6(self, request):
             return Response('Unexpected fields: %s' % ', '.join(extra), status=status.HTTP_400_BAD_REQUEST)
         if len(request.data) == 0:
             return Response('No fields specified for update.', status=status.HTTP_400_BAD_REQUEST)
+
+        from queue.models import QUEUE_ORDER_FIFO, QUEUE_ORDER_LIFO
+        if 'queue_mode' in request.data and request.data['queue_mode'].upper() not in [QUEUE_ORDER_FIFO, QUEUE_ORDER_LIFO]:
+            msg = 'Unexpected value %s for queue_mode. Valid values are %s, %s' % (request.data['queue_mode'],
```
gisjedi: 👏 Thanks for providing a good error with the 400 status code.

scale/scheduler/scheduling/manager.py
emimaesmith: I believe it adds jobs one at a time, but I'm not opposed to changing it to `>=`.

PR #1834, opened 2019-12-19
#1835 Return Ingest Status By Given Started/End Dates

emimaesmith:
##### Checklist

- [x] `manage.py test` passes
- [x] tests are included

### Affected app(s)

ingest

### Description of change

Updated the logic that returned the ingest status to only include values from the given time range instead of the beginning of the start day and end of the end day.

### DIFF
--- a/scale/ingest/views.py
+++ b/scale/ingest/views.py
@@ -3,9 +3,11 @@ from __future__ import unicode_literals
 
 import datetime
 import logging
+import json
 
 import rest_framework.status as status
 from rest_framework.renderers import JSONRenderer
+from django.http import JsonResponse
 from django.http.response import Http404
 import django.utils.timezone as timezone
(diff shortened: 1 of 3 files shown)


scale/ingest/views.py
```diff
@@ -179,8 +181,7 @@ def list_impl(self, request):
         :rtype: :class:`rest_framework.response.Response`
         :returns: the HTTP response to send back to the user
         """
-
-        started = rest_util.parse_timestamp(request, 'started', rest_util.get_relative_days(7))
+        started = rest_util.parse_timestamp(request, 'started', required=False)
```
Fizz11: Are we worried about a malformed date/time string in the request?

emimaesmith: This is handled in the the rest_util function call.

Fizz11: Looks good then.

PR #1846, opened 2020-02-03
Scheduler Logic Updates

emimaesmith:
##### Checklist

- [x] `manage.py test` passes
- [x] tests are included

### Affected app(s)

scheduler

### Description of change

Tweaked scheduler logic to fill QUEUE_LIMIT instead of only grabbing the first 500 items in the queue table as this could lead to constantly grabbing the same 500 unschedulable jobs.

#1851


emimaesmith:
Closing as I'm not convinced this is the right approach. Will revisit if future testing reveals we need to re-write the scheduler logic.

PR #1875, opened 2020-04-07
Cooling off on the logging

emimaesmith:
##### Checklist

- [x] `manage.py test` passes

### Affected app(s)

scheduler 

### Description of change

Removed the warnings on every pass of the process queue loop in favor of a "[count] [job_type_name] job types could not be scheduled due to [reason]" warning at the end of the processing loop. scale-ingests exceeding their scheduling limit were overrunning the output logs.

### DIFF
--- a/scale/scheduler/scheduling/manager.py
+++ b/scale/scheduler/scheduling/manager.py
@@ -374,7 +374,6 @@ class SchedulingManager(object):
                             }
                         continue
                     if resource.name not in max_cluster_resources._resources:
-                        # logger.warning('Job type %s could not be scheduled as resource %s does not exist in the available cluster resources', jt.name, resource.name)
                         if jt.name in type_warnings:
                             type_warnings[jt.name]['count'] += 1
                         else:
@@ -427,7 +426,6 @@ class SchedulingManager(object):
                             'warning': '%s job types could not be scheduled due to missing workspace' % jt.name,
                             'count': 1
                         }
-                    # logger.warning('Job type %s could not be scheduled due to missing workspace', jt.name)
                     continue
 
                 # Check limit for this execution's job type
--- a/scale/scheduler/scheduling/manager.py
+++ b/scale/scheduler/scheduling/manager.py
@@ -340,112 +340,113 @@ class SchedulingManager(object):
         type_warnings = {}
 
         # We can schedule as long as there are nodes
-        if nodes:
-            ignore_job_type_ids = self._calculate_job_types_to_ignore(job_types, job_type_limits)
-            max_cluster_resources = resource_mgr.get_max_available_resources()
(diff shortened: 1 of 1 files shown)


scale/scheduler/scheduling/manager.py
```diff
@@ -336,94 +336,122 @@ def _process_queue(self, nodes, job_types, job_type_limits, job_type_resources,
         """
 
         scheduled_job_executions = []
-        ignore_job_type_ids = self._calculate_job_types_to_ignore(job_types, job_type_limits)
         started = now()
+        type_warnings = {}
+
+        # We can schedule as long as there are nodes
+        if nodes:
```
gisjedi, line 343: I'm not a big fan of nesting the entire scheduler loop within this if block. 2 reasons:
1. It's really hard to review what actually changed now that every line is _changed_ due to whitspace.
2. Having the shortcut cases up front where the loop no-ops is nice for tracing logically in my head.

scale/scheduler/scheduling/manager.py
```diff
@@ -336,94 +336,122 @@ def _process_queue(self, nodes, job_types, job_type_limits, job_type_resources,
...
+                # Try to schedule job execution and adjust job type limit if needed
+                if self._schedule_new_job_exe(job_exe, nodes, job_type_resources):
+                    scheduled_job_executions.append(job_exe)
+                    if job_type_id in job_type_limits:
+                        job_type_limits[job_type_id] -= 1
+        else:
+            logger.warning('There are no nodes available. Waiting to schedule until there are free resources...')
 
         duration = now() - started
+        if type_warnings:
```
gisjedi, line 451: This will still spam on each pass of the scheduler loop. This might be enough for now, but in the future we might want to consider only outputting warnings once every 5 seconds or some such.

emimaesmith: It at least won't spam 450 lines out of the 500 loops (or whatever the queue limit is set to). Just one line for each job type after it loops through the queue to give you a summary of what jobs couldn't be scheduled that time around.

scale/scheduler/scheduling/manager.py
```diff
@@ -367,10 +368,22 @@ def _process_queue(self, nodes, job_types, job_type_limits, job_type_resources,
             for resource in job_exe.required_resources.resources:
                 # skip sharedmem
                 if resource.name.lower() == 'sharedmem':
```
gisjedi, line 370: `sharedmem` isn't a resource that is recognized by Mesos and therefore outputting on warning on it is meaningless. `sharedmem` is only a configuration that is offered by way of our `resources` job-type config that is passed into Docker as a special parameter.

gisjedi: That said the if block needs to remain so we skip marking `sharedmem` as a required resource and don't try to tell Mesos our job needs the non-existent `sharedmem` resource.

Issue #1885, opened 2020-06-09
Remove Unique File Name Requirement


emimaesmith:
**Pain Point? Please describe.**
Scale currently uses the file name to determine if a source file is a new, unique file or a duplicate of a file that's been previously processed. This is fine if we can guarantee all incoming data to have unique filenames, however this is not always the case. There's no reason other file metadata couldn't be used to assist in determining source file uniqueness instead of simply relying on the filename.

**Desired Solution**
Rather than only using the filename to check for file uniqueness, a combination of at least the file size and file name should be used when checking a source file to see if it is a duplicate. At the point of ingest, this information (in addition to media type, file path, and additional tags) are already known about the file.

From an ingest perspective, there is no reason why filename should be the only part of the file checked for uniqueness - the file_name database column is not currently marked as the PK or required to be unique for the ScaleFile table. To verify this change will not cause any negative effects, we need to ensure other aspects of the codebase (job data creation, file retrieval, etc) do not also count on this file name to be a unique field.

PR #1886, opened 2020-06-18
#1885 Remove Unique Filenames requirement

emimaesmith:
##### Checklist

- [x] `manage.py test` passes
- [x] tests are included

### Affected app(s)

ingest

### Description of change

#1885 
Updated the ingest (strikes and scans) to check for duplicate files by filename and file size instead of just filename.

I think this is ready - we might need some more testing with S3 strikes but from the code as far as I can tell we should be good.

### DIFF
--- a/scale/ingest/strike/monitors/dir_monitor.py
+++ b/scale/ingest/strike/monitors/dir_monitor.py
@@ -269,7 +269,9 @@ class DirWatcherMonitor(Monitor):
                 logger.info('Transfer complete: %s', file_path)
 
         if ingest.status == 'TRANSFERRED':
-            ingest_path = os.path.join(self._ingest_dir, file_name)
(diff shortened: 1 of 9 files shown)


scale/ingest/strike/monitors/dir_monitor.py
```diff
@@ -269,7 +269,8 @@ def _process_file(self, file_name, ingest):
                 logger.info('Transfer complete: %s', file_path)
 
         if ingest.status == 'TRANSFERRED':
-            ingest_path = os.path.join(self._ingest_dir, file_name)
+            # TODO: do the same for S3?
```
Fizz11, line 272: should probably verify one way or another

emimaesmith: Couldn't find any evidence S3 would have an issue. Removed the TODO.

Issue #1883, opened 2020-06-01
Basic Scale job how-to


Fizz11:
While we have documentation for most parts of scale, it would be beneficial to have a start to finish tutorial of creating a simple job and incorporating it in scale. Instructions for the following tasks should be written out in a simple, friendly blog-style post with screenshots and code snippets where applicable. This tutorial will assume the scale infrastructure is already up and running. 

1. Create a simple program. 
    - program should have one input and one output. 
    - program should take two arguments, 'INPUT_FILE' which will be the filepath for the input file, and "OUTPUT' which will be the directory path for the output directory. 
    - program should copy input file with a new name and write it to the output directory. 

2. Create a seed manifest file for the program
    - documentation here: https://ngageoint.github.io/seed/seed.html
    - the seed-cli app can be used to verify the seed.manifest.json file. https://github.com/ngageoint/seed-cli

3. dockerize program and build with seed-cli
    - create a dockerfile and base the image on alpine
    - build image using seed-cli command `seed build`
    - upload image to docker hub

4. Add new image to seed-silo
seed-silo documentation here: https://github.com/ngageoint/seed-silo#usage

5. set up a new workspace in scale-ui
this will be a local workspace pointing at a shared folder that all scale nodes have access to. 

6. create job-type for program in scale-ui

7. create new recipe for job-type in scale-ui

8. create new strike in scale-ui

9. add files to input workspace and verify recipe and job run correctly

PR #1887, opened 2020-06-23
Scale Walkthrough Documentation

mattbanderson:
#1883 - Basic Scale job how-to

##### Checklist

### Affected app(s)
Documentation

### Description of change
Add Scale walkthrough documentation to web docs. Incorporate AsciiDoc HTML generation into Travis CI build process. Update tutorials layout since there are now four instead of three.


web-docs/walkthrough/index.adoc
```diff
@@ -0,0 +1,369 @@
...
+ENTRYPOINT ["python", "input-output.py"]
+----
+
+The Dockerfile defined in this example takes the following steps:
+
+* Start `FROM` the pre-existing `alpine:3.9` Linux image. This is an official image, validated by Docker.
+* `RUN` the command `apk -U add python` inside your image filesystem, which will install Python.
+* `COPY` the file input-output.py to your work directory.
+* Use `WORKDIR` to create the `/app` directory and specify that all subsequent actions should be taken from within `/app` _in your image filesystem_ (never the host’s filesystem).
+* Use `ENTRYPOINT` to describe how to run the container. In this case, it will run `python input-output.py`. The arguments will be specified at runtime.
```
gisjedi, line 126: I'd say we omit use of entry point as we did in seed guide. The use of seed "command" should be preferred.

web-docs/walkthrough/index.adoc
```diff
@@ -0,0 +1,369 @@
...
+      "output"
+    ],
+    "maintainer": {
+      "name": "Nikkala Thomson",
+      "organization": "AIS",
+      "email": "nikkala.thomson@appliedis.com"
+    },
+    "timeout": 3600,
+    "interface": {
+      "command": "${INPUT_FILE} ${OUTPUT_DIR}",
```
gisjedi, line 164: Related to the entry point comment from above we should put the executable in here.

web-docs/walkthrough/index.adoc
```diff
@@ -0,0 +1,369 @@
...
+image:job-type-general-info.png[]
+. Click on the “4” (Validate and Create), click “Validate” and then “Save”. This will create your job type.
+
+=== Create a Recipe to Run an Instance of a Job Type
+
+A recipe is a collection of one or more job types that interact with each other in a workflow.
+
+. Navigate to “Configuration -> Recipe Types” then click on the “+” to create a new recipe.
+. Enter a title for your recipe, and a description if you like.
+. Click on “File Inputs” in the middle right and enter a name (commonly “input_file”) to represent your input file, set “Required” to “True,” then click “+ Add File to Input” and close the window.
```
gisjedi, line 322: Due to case mismatch and challenges we've had when not set to all caps. I'd just recommend we always prefer caps.

web-docs/walkthrough/index.adoc
```diff
@@ -0,0 +1,369 @@
...
+:docker-hub-signup-url: https://hub.docker.com/signup
+:docker-reference-url: https://docs.docker.com/engine/reference/builder 
+:scale-github-url: https://github.com/ngageoint/scale
+:scale-gh-pages-url: https://ngageoint.github.io/scale
+:seed-url: https://github.com/ngageoint/seed
+:seed-cli-url: https://github.com/ngageoint/seed-cli
+:seed-install-url: https://ngageoint.github.io/seed-cli/#_seed_cli_installation
+:seed-releases-url: https://github.com/ngageoint/seed-cli/releases
+:seed-silo-url: https://github.com/ngageoint/seed-silo
+:seed-spec-url: https://ngageoint.github.io/seed/seed
```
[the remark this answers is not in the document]
mattbanderson: Which URL?

web-docs/walkthrough/index.adoc
mattbanderson: Good call, pushed an update to that uses `INPUT_FILE` as the example.

web-docs/walkthrough/index.adoc
```diff
@@ -0,0 +1,369 @@
...
+You will also need an input file in the right format for your process. 
+For this Python process, create a simple text file with a few lines of text, and name it with the `.txt` extension. 
+Save it in the same folder as your process.
+
+=== Download and Install the Seed Command Line Interface (CLI)
+
+You will find the latest version of Seed here: {seed-releases-url}
+
+* Scroll down and click on the green-outlined button on the left that says “Latest Release”
+* Scroll down to “Assets” and download the Seed version for your operating system. Note: “darwin” is the correct version for macOS.
```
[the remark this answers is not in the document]
mattbanderson: I think the context behind this note is Nikkala going through the process as a very new Scale user. I pushed an update that uses a 'tip' to clearly delineate which Seed release binary to use for each OS (even if that may be obvious to you and I).

web-docs/walkthrough/index.adoc
mattbanderson: I agree with you that it would be ideal to have this example match what's in the Seed user guide. 

That said, I simply took what Nikkala provided, converted it to AsciiDoc, made a few formatting changes, and incorporated it into the existing Scale docs.

I would suggest we merge this as-is, which will validate whether we've successfully integrated AsciiDoc into the existing documentation structure and add an issue to update the example to more closely match what's in the Seed user guide. 

With AsciiDoc successful integrated into the CI build, anyone should be able to update the docs going forward once they learn some AsciiDoc syntax, since they will have an example to work from and not need to worry about how the build works.

Thoughts? If you feel strongly we should update the example before merging this PR, I can update it or kick it back to Nikkala to have her make the changes.

web-docs/walkthrough/index.adoc
mattbanderson: See above.

web-docs/walkthrough/index.adoc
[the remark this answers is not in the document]
gisjedi: https://ngageoint.github.io/seed/seed -> https://ngageoint.github.io/seed/seed.html

Surprisingly enough both work, but the later is the canonical most recent version page.

web-docs/walkthrough/index.adoc
gisjedi: I'm in agreement. Let's create an issue to fix this after merge.

web-docs/walkthrough/index.adoc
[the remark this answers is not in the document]
mattbanderson: Updated the URL in [2df6f0c](https://github.com/ngageoint/scale/pull/1887/commits/2df6f0cdd5d65efcc43be8cd5cbd1932def0ea9d)

web-docs/walkthrough/index.adoc
mattbanderson: Created issue [1890](https://github.com/ngageoint/scale/issues/1890)

Issue #1891, opened 2020-06-30
CI build process ignores AsciiDoc output


mattbanderson:
**Description**
The CI build process for Scale docs generates the documentation, adds the generated files, and commits them to the GitHub pages branch to update the published documentation. The output of the AsciiDoc document generation is not being added to the GitHub pages branch because it has a `.gitignore` file to exclude the contents of the output directory. This is intended to avoid committing the output directory when developing locally but is causing the documentation generated by the CI build to not get updated.

**Reproduction Steps**
Steps to reproduce the problem:
1. Start a Travis CI build
2. Go to the build output
3. At the end of the build output, observe that no files under the `output` directory are added

**Expected behavior**
Output generated by the AsciiDoc build should be added to the published documentation.

**Version and Environment Details**
 - [GitHub Travis CI build](https://travis-ci.org/github/ngageoint/scale)

PR #1892, opened 2020-06-30
Fix AsciiDoc generation in Travis CI build

mattbanderson:
#1891 - CI build process ignores AsciiDoc output

##### Checklist

- [x ] documentation is changed or added

### Affected app(s)
Documentation

### Description of change
Move contents of `.gitignore` in walkthrough directory to repo-level `.gitignore` to fix inadvertent exclusion of AsciiDoc output from published documentation.


mattbanderson:
Reasoning my way through the CI build scripts, I _think_ removing the `.gitignore` file from the walkthrough directory will solve the issue. I moved the contents to the repo-wide `.gitignore`, so `output` will still be excluded from inadvertently committed for local dev. Since only files under `scale/docs` and `web_docs` get copied into the staging directory and pushed to the `gh-pages` branch, I think this solves the issue.

PR #1903, opened 2020-07-23
Moved value to environment variable.

loganr-w:
### Checklist

- [x] `manage.py test` passes

### Affected app(s)

Scale Scheduler

### Description of change

Moved the hard-coded value for max connections in the scheduler to an environment variable.

The hope is that by increasing this number that when a leader election happens that takes an extended amount of time, the framework will not be marked inactive.


loganr-w:
Setting `SCHEDULER_MAX_RECONNECT` to a higher number (say 30) should fix the issue where the framework is going inactive. We set it to 30 and trigged a leader election several times and the framework did not going inactive. We set it to 1, triggered a leader election and the framework went inactive. Hopefully it works.

